Issue #3, opened 2024-02-19
Optimistic updates


posva:
Add methods to mutations to handle optimistic data updates

ElisePatrikainen:
Maybe a first step could be to add some hooks (such as `onMutate`, `onError`, `onSuccess`, `onSettled`)?

PR #12, opened 2024-03-12
feat(useMutation): add hooks

ElisePatrikainen:
Add hooks to enable optimistic updates (relates to https://github.com/posva/pinia-colada/issues/3).

Which raises now few questions:
- should the hooks be awaited in the `mutate` function?
- hook arguments:
   - should the `mutate` function arguments be passed to hooks? So far, I have passed them only to the `onMutate` one.
   - would passing a context (where the user can store any data) between hook be interesting?
   - `onSettled`: should it get the promise result as arguments (the data if the promise succeeded, or the error if it failed)?
- in the case where the `onError` hook is provided by the user, should, in case of errors, an error still be thrown? Or should we throw it only if the `onError` hook is not provided?


playground/db.json
```diff
@@ -58,15 +58,15 @@
       "color": "Sienna",
       "imagesrc": "https://tailwindui.com/img/ecommerce-images/shopping-cart-page-01-product-01.jpg",
       "price": 24,
-      "availability": -2
+      "availability": 1
```
posva, line 61: This should be reverted 😄

playground/src/api/products.ts
```diff
@@ -26,7 +27,8 @@ export function getProductById(id: string | number, options?: Options<'json'>) {
   return products.get<ProductT>(id, options)
 }
 
-export async function changeProductAvailability(product: ProductListItem, options?: Options<'json'>) {
+export async function changeProductAvailability(product: ProductListItem, options?: Options<'json'>, _delay?: number) {
+  await delay(_delay ?? 0)
```
posva: let's not add the delay as an argument. We can always add it later to the inner function itself (randomized delay) or within the mutation fn

src/use-mutation.ts
```diff
@@ -23,6 +23,26 @@ export interface UseMutationOptions<
    */
   keys?: _MutationKeys<TParams, TResult>
 
+  /**
+   * Hook to execute a callback when the mutation is triggered
+   */
+  onMutate?: (...args: TParams) => void
```
posva: It should be able to return a context (or a promise of it) that is later passed to other hooks

```suggestion
  onMutate?: (...args: TParams) => Promise<TContext> | TContext
```

src/use-mutation.ts
```diff
@@ -23,6 +23,26 @@ export interface UseMutationOptions<
...
+
+  /**
+   * Hook to execute a callback in case of error
+   */
+  onError?: (error: any) => void
```
posva: This should get access to the arguments and the error. I feel like at this point it's better to just have one object:

```suggestion
  onError?: (context: { error: TError, args: TParams } & TContext) => Promise<unknown> | unknown
```

It does have a risk of collision but it should be fine and easier to use than

```suggestion
  onError?: (error: TError, context: TContext, ...args: TParams) => void
```

src/use-mutation.ts
```diff
@@ -23,6 +23,26 @@ export interface UseMutationOptions<
...
+
+  /**
+   * Hook to execute a callback in case of error
+   */
+  onSuccess?: (result: TResult) => void
```
posva: Same as error

src/use-mutation.ts
```diff
@@ -23,6 +23,26 @@ export interface UseMutationOptions<
...
+
+  /**
+   * Hook to execute a callback in case of error
+   */
+  onSettled?: () => void
```
posva: This one should get access to both error and data

src/use-mutation.ts
```diff
@@ -85,10 +105,17 @@ export function useMutation<
...
+      options.onMutate(...args)
+    }
+
     // TODO: AbortSignal that is aborted when the mutation is called again so we can throw in pending
     const promise = (pendingPromise = options
       .mutation(...args)
       .then((_data) => {
+        if (options.onSuccess) {
+          options.onSuccess(_data)
+        }
```
posva:
```suggestion
        options.onSuccess?.(_data)
```

src/use-mutation.ts
```diff
@@ -111,7 +138,15 @@ export function useMutation<
           error.value = _error
           status.value = 'error'
         }
+        if (options.onError) {
+          options.onError(_error)
+        }
```
posva: same as above

src/use-mutation.ts
```diff
@@ -111,7 +138,15 @@ export function useMutation<
...
         throw _error
+      })
+      .finally(async () => {
+        if (options.onSettled) {
+          options.onSettled()
+        }
```
posva: same

posva (changes requested):

PR #14, opened 2024-03-17
docs: queries guide

ElisePatrikainen:


docs/guide/queries.md
```diff
@@ -1 +1,71 @@
 # Queries
+
+## Definition:
```
posva: let's drop this, it's not needed, we can directly state what queries are

docs/guide/queries.md
```diff
@@ -1 +1,71 @@
...
+
+Queries are dependencies to an async source of data. They allow us to declaratively fetch, cache and refresh the data.
+
+::: info
+Queries are meant to **read** data. In terms of REST, for example, queries would handle `GET` requests. If you need to **write** (or mutate) data, you can use [mutations](./mutations.md).
+:::
```
posva: No need to put this is an info as it doesn't go off topic

docs/guide/queries.md
```diff
@@ -1 +1,71 @@
...
+
+## Defining a query:
```
posva: Same as above, it should be introduced directly.

docs/guide/queries.md
```diff
@@ -1 +1,71 @@
...
+
+Alternatively, a query can be defined with the `defineQuery()` function, which allows you to create related properties associated with the queries.
+
+// TODO
+
+## Query options and return:
+
+// TODO: link mentioned options to the API doc
+
+### Options:
```
posva:
```suggestion
### Options

```

docs/guide/queries.md
```diff
@@ -1 +1,71 @@
...
+- [`key`](./query-keys.md): the key used internally to identify the query. It can also be used to [invalidate a query](./query-invalidation.md).
+- `query`: the function called to fetch the data, which **must** be async.
+
+The other options are not required. They are related to:
+- the cache: for example, the `staleTime` option, which defines the time after which the data is considered stale, or the `gcTime`, which defines the time after which an inactive query should be garbage collected.
+- revalidation events: the query cache can be revalidated on specific events (component mount, window focus or reconnection). We can then, for each of these events, define whether the data should be revalidated or not, and how (refreshed or refetch, cf the 'returns' section for more details // TODO: add link).
+
+// TODO: how to define options defaults,
+
+### Returns:
```
posva: There shouldn't be a section named return, it looks more like an API than a guide. It sohuld be about what can I do with the returned values of `useMutation()`

docs/guide/queries.md
```diff
@@ -1 +1,71 @@
 # Queries
+
+## Definition:
+
+Queries are dependencies to an async source of data. They allow us to declaratively fetch, cache and refresh the data.
```
posva: It should be more along the lines of

```suggestion
Queries handle async state declaratively. They let you focus on the state, its async status and any eventual error. They also automatically dedupe multiple requests and cache results to create a fast user experience.
```

docs/guide/queries.md
```diff
@@ -1 +1,71 @@
...
+
+const { data, error, isFetching } = useQuery({
+  key: ['todos'],
+  query: () => fetch('/api/todos').then((res) => res.json())
+})
+</script>
+
+<template>
+  <main>
+    <div v-if="isFetching">
```
posva: let's use the `status` instead, it should be the recommended approach as it covers more cases while `isFetching` is just a convenient wrapper

docs/guide/queries.md
```diff
@@ -1 +1,71 @@
...
+      Loading
+    </div>
+    <div v-else-if="error">
+      Oups, an error happened...
```
posva:
```suggestion
      Oops, an error happened...
```

docs/guide/queries.md
```diff
@@ -1 +1,71 @@
...
+    </div>
+    <div v-else>
+      <Todo v-for="todo in data" :key="todo.id" :todo />
```
posva:
```suggestion
      <TodoItem v-for="todo in data" :key="todo.id" :todo />
```

docs/guide/queries.md
```diff
@@ -1 +1,71 @@
...
+    </div>
+  </main>
+</template>
+```
+
+This composable:
+- accepts an option object, which configures the key of the query, how to fetch de data, and options related to the cache and its revalidation
+- returns the state of the query, which can be used in the UI, and some methods to declaratively revalidate the cache if needed
```
posva: Let's cover the basics first: the query, then the key, what is expected from them. Then explain how the query is automatically triggered when needed (no need to go into details yet) and that it will update `data`, and others enabling a declarative approach

docs/guide/queries.md
```diff
@@ -1 +1,71 @@
...
+// TODO
+
+## Query options and return:
+
+// TODO: link mentioned options to the API doc
+
+### Options:
+Queries have two mandatory options:
+- [`key`](./query-keys.md): the key used internally to identify the query. It can also be used to [invalidate a query](./query-invalidation.md).
+- `query`: the function called to fetch the data, which **must** be async.
```
posva: it must return a promise (more correct)

docs/guide/queries.md
```diff
@@ -1 +1,71 @@
...
+
+The other options are not required. They are related to:
+- the cache: for example, the `staleTime` option, which defines the time after which the data is considered stale, or the `gcTime`, which defines the time after which an inactive query should be garbage collected.
+- revalidation events: the query cache can be revalidated on specific events (component mount, window focus or reconnection). We can then, for each of these events, define whether the data should be revalidated or not, and how (refreshed or refetch, cf the 'returns' section for more details // TODO: add link).
```
posva: This should go into different sections that cover a more high-level pov of the option by giving some concrete usage and problem solving. e.g. 

> The query will automatically consider the data _fresh_ for 5 seconds. Only after 5 seconds, the data will be marked as _stale_ and any new `useQuery()` or `refresh()` calls with the same key will execute the `query` again.

posva (changes requested):

docs/guide/queries.md
```diff
@@ -1 +1,88 @@
...
+      <TodoItem v-for="todo in data" :key="todo.id" :todo />
+    </div>
+  </main>
+</template>
+```
+
+// TODO: link mentioned options/returns to the API doc
+
+As we can see, to define a query, we need at least:
+- a key: it will be used to store and retrieve the query
```
posva:
```suggestion

- A unique `key`. It allows to reuse of any query 
```

docs/guide/queries.md
```diff
@@ -1 +1,88 @@
...
+    <div v-else>
+      <TodoItem v-for="todo in data" :key="todo.id" :todo />
+    </div>
+  </main>
+</template>
+```
+
+// TODO: link mentioned options/returns to the API doc
+
+As we can see, to define a query, we need at least:
```
posva:
```suggestion
All queries require two properties:
```

docs/guide/queries.md
```diff
@@ -1 +1,88 @@
...
+- a key: it will be used to store and retrieve the query
+- the query function: the actual function to fetch the data (indeed, Pinia Colada does not and is not meant to provide an HTTP client, and is therefore agnostic on the way that you fetch the data, the only constraint being that the query function returns a promise)
```
posva:
```suggestion
- A `query` function. The actual function to fetch the data (indeed, Pinia Colada does not and is not meant to provide an HTTP client and is therefore agnostic on the way that you fetch the data, the only constraint being that the query function must return a promise)
```

docs/guide/queries.md
```diff
@@ -1 +1,88 @@
...
+`useQuery` accepts other options, which are not required, to configure the query cache (for example, the `staleTime`) or when and how the query should be triggered.
```
posva:
```suggestion

`useQuery` accepts other options to configure the query cache (for example, the `staleTime`) or when and how the query should be triggered.
```

docs/guide/queries.md
```diff
@@ -1 +1,88 @@
...
+
+Then, the query will be automatically triggered when needed (more precisely on specific events cf // TODO: link to the relevant section), either to fetch or to refresh the data, enabling a declarative approach. You can access to the fetched data through the references returned by the composable (`data`), as well as to more information on the query state, such as its status (`status`). It also returns methods to manually trigger the query if needed.
```
posva:
```suggestion
Queries are automatically triggered when needed (more precisely on specific events cf // TODO: link to the relevant section), **enabling a declarative approach**. You can access the fetched data through the references returned by the composable (`data`), as well as other query state, such as its `status`, `error` and more. It also returns methods to manually trigger the query.
```

docs/guide/queries.md
```diff
@@ -1 +1,88 @@
...
+
+## The `defineQuery()` API:
```
posva:
```suggestion
## Reusable Queries
```

docs/guide/queries.md
```diff
@@ -1 +1,88 @@
...
+
+Alternatively, a query can be defined with the `defineQuery()` function, which allows you to create related properties associated with the queries. This is particularily usefull if we need to share a query between components.
```
posva:
```suggestion
While `useQuery()` can be directly called in components, we often need to reuse the same query across components or even add extra properties like consuming the route via `useRoute()` or passing a _search_ text to the API request. In those scenarios, it's convenient to _define_ queries and reuse them where needed. 
```

docs/guide/queries.md
```diff
@@ -1 +1,88 @@
...
+
+Let's take an example, where we have a query depending on a search param. Let's assume we want to abstract a query into a composable:
+
+```ts twoslash
+import { useQuery } from '@pinia/colada'
```
posva, line 53:
```suggestion
// src/queries/todos.ts
import { useQuery } from '@pinia/colada'
```

docs/guide/queries.md
```diff
@@ -1 +1,88 @@
...
+import { ref } from 'vue'
+
+export const useSearchedTodos = () => {
```
posva:
```suggestion
export function useFilteredTodos() {
```

docs/guide/queries.md
```diff
@@ -1 +1,88 @@
...
+`useQuery` accepts other options, which are not required, to configure the query cache (for example, the `staleTime`) or when and how the query should be triggered.
+
+Then, the query will be automatically triggered when needed (more precisely on specific events cf // TODO: link to the relevant section), either to fetch or to refresh the data, enabling a declarative approach. You can access to the fetched data through the references returned by the composable (`data`), as well as to more information on the query state, such as its status (`status`). It also returns methods to manually trigger the query if needed.
+
+## The `defineQuery()` API:
+
+Alternatively, a query can be defined with the `defineQuery()` function, which allows you to create related properties associated with the queries. This is particularily usefull if we need to share a query between components.
+
+Let's take an example, where we have a query depending on a search param. Let's assume we want to abstract a query into a composable:
```
posva:
```suggestion
You might think that we could just create a regular composable for this:
```

docs/guide/queries.md
```diff
@@ -1 +1,88 @@
...
+  const query = useQuery({
+    key: () => ['todos', search.value],
+    query: () =>
+      fetch(`/api/todos?search=${search.value}`, { method: 'GET' }),
+  })
+  return { search, ...query }
+}
+```
+
+This implementation presents couple of drawbacks.
```
posva:
```suggestion
But this implementation has a few drawbacks.
```

posva (commented):
I will merge and apply some extra changes locally

Issue #11, opened 2024-03-06
A way to avoid using outside state


posva:
maybe a `defineQuery(() => {})` to allow using globals and creating global state associated with the query (e.g. a page?).

Currently one can try to do this:

```vue
<script setup lang="ts">
const page = ref(1)
useQuery({
  query() {
     return fetchData(page.value)
  },
  key: () => ['key', page.value]
})
<script>
```

But this can fail in _some_ scenarios, notably if the same key is used in a different component or if this gets abstracted with

```ts
export function useSomeData() {
	const page = ref(1)
	return useQuery({
	  query() {
	     return fetchData(page.value)
	  },
	  key: () => ['key', page.value]
	})
}
```

and used in multiple components, the `page` will end up scoped to the first component that calls it. Both components will create a page ref but only the first one will be used within query and key. Therefore, if the first component is unmounted while the other still lives, the page ref will not be reactive anymore, breaking the usage.

### Notes

- An eslint plugin could warn agains this usage. It would be better to have some runtime warning but not sure how to do it, maybe checking for the active scope and saving it somewhere in dev only?
- ~~[Maybe run within a computed and check its deps](https://play.vuejs.org/#eNp9kk9Lw0AQxb/KuJdUqAmip5IWVArqQUUFL3vJn0mautlddjexEPLdnU1MW1R625n3m+G9YTt2o3XYNsgWLLaZqbQDi67RKy6rWivjoAODxRwyVevGYQ49FEbVENBQwCWXmZLWQW1LWHpyFtyjEAo+lBH5WXA+AQnJ047Z7ByWKwjwIKd/ZVoZtolokKBkfHGZTg8/pgSGQpWzIIUctQ3mkIZYFJi50Nc/y/dUMlHJLyqOxugUmgqHtRaJQ6oA4s3lquuGeH0fR1QN3UqSU2gvapWjWHJGOmcQkRhHR/NszpwlD0VVhlurJF258/Oc+ayVQPOsXUUeOVvAoHgtoft9PQ49ZxqcT/1sg9nnP/2t3fkeZy8GLZoWOdtrLjElulFevz3hjt57kdw3gugT4ivS/RrvccRuG5mT7SNucPsw/JVKlu92vXMo7RTKG/VkP/Cc0a+5OxH9YPcqvB7muOxZ/w3S5OGn)~~ no longer valid in latest vue
- The function version is nice to setup initial state. Each query/mutation is effectively a mini store

posva:
[...]

PR #19, opened 2024-03-21
feat(mutation): add an optional `key` option to mutations

ElisePatrikainen:


src/entry-options.ts
```diff
@@ -0,0 +1,7 @@
+import type { EntryNodeKey } from './tree-map'
+import type { _ObjectFlat } from './utils'
+
+/**
+ * Key used to identify a query. Always an array.
```
posva: Let's adapt the description

src/entry-options.ts
```diff
@@ -0,0 +1,7 @@
...
+ */
+export type UseEntryKey = Array<EntryNodeKey | _ObjectFlat>
```
posva: Since it's not related to `use...` anymore. Let's just name it EntryKey

src/use-mutation.ts
```diff
@@ -55,6 +55,8 @@ export interface UseMutationOptions<
    */
   mutation: (vars: TVars, context: NoInfer<TContext>) => Promise<TResult>
 
+  key?: MaybeRefOrGetter<UseEntryKey>
```
posva: In mutations, it should be able to access the variables passed to the mutation, similar to the invalidate keys

I'm realizing that allowing a computed property or a ref is actually problematic as it could create the issues highlighted in #11. It's something we will have to look into later on

ElisePatrikainen: I am wondering, for the ref or computed properties, is it then really worth handling them? I am just realising that users would probably rather trigger a mutation imperatively than declaratively relying on reactivity, and therefore I am wondering if passing reactivity through the `key` (and the issue associated) is enough profitable. Maybe I could just type the `key` the same way as you did for the `keys` property (without reactivity).

posva: You are right; passing a ref or a computer seems like a foot-gun. It should only be static or computed from the arguments

src/use-mutation.ts
```diff
@@ -69,6 +73,8 @@ export interface UseMutationOptions<
    */
   mutation: (vars: TVars, context: NoInfer<TContext>) => Promise<TResult>
 
+  key?: _MutationKey<TVars, TResult>
```
posva: this key should happen before onMutate so it will only have access to `vars`.

posva (changes requested):

PR #29, opened 2024-04-18
feat: add gcTime

ElisePatrikainen:


ElisePatrikainen:
@posva I initially thought of tracking the instances where queries are used, and triggering the gc timeout when all instances are unmounted (ie when the query becomes inactive) but I didn’t find a way to track it. Therefore, as a first step, I am triggering it at each fetch (which, I think, could at least cover a consistent part of the cases, because it seems quite likely that, in the case of an active query, it would be refetch before the completion of the gcTime).

src/query-store.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+    enableAutoUnmount(afterEach)
+
+  it('should delete the cache of an inactive query once the gcTime is elapsed', async () => {
+    const useTodoList = defineQuery({
+      key: ['todos'],
+      query: async () => [{ id: 1 }],
+      gcTime: 1000,
+    })
+
+    let queryCache: any // TODO: improve
```
posva:
```suggestion
    let queryCache!: ReturnType<typeof useQueryCache>
```

src/query-store.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+          queryCache = useQueryCache()
+        },
+      },
+      {
+        global: {
+          plugins: [createPinia(), QueryPlugin],
+        },
+      },
+    )
+    expect(queryCache.caches.get(['todos']).gcTimeoutId).toBeDefined()
```
posva: This is an implementation detail, so it should be removed
```suggestion
```

src/query-store.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+    vi.advanceTimersByTime(500)
```
posva: What about doing the exact values?
```suggestion
    vi.advanceTimersByTime(999)
```

src/query-store.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+    expect(queryCache.caches.get(['todos'])).toBeDefined()
+    vi.advanceTimersByTime(1200)
```
posva:
```suggestion
    vi.advanceTimersByTime(1)
```

src/query-store.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+
+    expect(queryCache.caches.get(['todos']).gcTimeoutId).toBeDefined()
+    vi.advanceTimersByTime(500)
+    expect(queryCache.caches.get(['todos'])).toBeDefined()
+
+    await queryCache.refresh(queryCache.caches.get(['todos']))
+    vi.advanceTimersByTime(999)
+    expect(queryCache.caches.get(['todos']).gcTimeoutId).toBeDefined()
+    vi.advanceTimersByTime(1)
+    expect(queryCache.caches.get(['todos'])).toBeUndefined()
```
posva: I think here it's really important to use the correct timing:

```suggestion
    vi.advanceTimersByTime(999)
    expect(queryCache.caches.get(['todos'])).toBeDefined()

    await queryCache.refresh(queryCache.caches.get(['todos']))
    vi.advanceTimersByTime(1)
    expect(queryCache.caches.get(['todos'])).toBeDefined()
		vi.advanceTimersByTime(1000)
		expect(queryCache.caches.get(['todos'])).toBeUndefined()
```

PR #35, opened 2024-04-28
feat(query): add `useQueriesState` composable

ElisePatrikainen:


src/query-store.ts
```diff
@@ -173,7 +173,7 @@ export const useQueryCache = defineStore(QUERY_STORE_ID, () => {
 
   function ensureEntry<TResult = unknown, TError = ErrorDefault>(
     keyRaw: EntryKey,
-    options: UseQueryOptionsWithDefaults<TResult, TError>,
+    options?: UseQueryOptionsWithDefaults<TResult, TError>,
```
quiteeasy, line 176: Creating a query entry without options risks errors if the key is invalidated and the `options` are accessed. Probably, we should consider avoiding creating query entities in `UseQueryState` and return `undefined/null` instead.

src/use-queries-state.ts
```diff
@@ -0,0 +1,31 @@
...
+
+// TODO: accept also a since key (instead of an array) as argument
+// TODO: accept a ref for `keys` (to handle adding/removing entries for example)?
+export function useQueriesState(keys: Array<MaybeRefOrGetter<EntryKey>>) {
+  const { ensureEntry } = useQueryCache()
+  return computed(() => {
+  return keys.map<QueryState[]>(
+    // TODO: fix TS issue
+    (key) => {
+        const query = ensureEntry(toValue(key))
```
quiteeasy, line 21: Should we add and use the `findQuery` function here? It will work if we decide not to create a query in `useQueriesState` in case it has not been created

src/use-queries-state.ts
```diff
@@ -0,0 +1,31 @@
+import type { MaybeRefOrGetter } from 'vue'
+import { computed, toValue } from 'vue'
+import type { EntryKey } from './entry-options'
+import type { QueryStatus } from './query-store'
+import { useQueryCache } from './query-store'
+
+interface QueryState {
```
quiteeasy, line 7: Can we use `_UseQueryEntry_State` type from `query-store.ts` here?

posva:
This is now available as `useQueryState` 🎉

PR #43, opened 2024-05-16
feat(use-query): Add enabled option

### LINKED ISSUE #42: Implement `enabled` option in `useQuery()` to control the automatic querying
This will controll the initial fetching and also any automatic refetching like refetchOnFocus, refetchOnMount. There should be a way for this to be reachable by plugins so a `refetchInterval` can be implemented but still controlled by this option.
It should allow a ref or getter too and immediately refresh if going from false to true

quiteeasy:
Close #42 

I have opened this PR to introduce an 'enable' option for the `useQuery` feature. This option will allow users to toggle the `useQuery` refetch/refresh internal behavior on or off based on their requirements.

I have added unit tests to ensure the new option works as expected and I will update the documentation accordingly after merging.

Please review the changes and provide any feedback or suggestions you may have.

### DIFF
--- a/src/query-options.ts
+++ b/src/query-options.ts
@@ -1,4 +1,4 @@
-import { type InjectionKey, type MaybeRefOrGetter, inject } from 'vue'
+import { type InjectionKey, type MaybeRefOrGetter, type WatchSource, inject } from 'vue'
 import type { EntryKey } from './entry-options'
 import type { QueryPluginOptions } from './query-plugin'
 import type { ErrorDefault } from './types-extension'
@@ -61,6 +61,12 @@ export interface UseQueryOptions<TResult = unknown, TError = ErrorDefault> {
   query: (context: UseQueryFnContext) => Promise<TResult>
 
   /**
+   * Reactive boolean, indicating whether refetch/refresh should be called internally,
+   * calling refetch when toggled to true
+   */
+  enabled?: WatchSource<boolean>
+
+  /**
    * Time in ms after which the data is considered stale and will be refreshed on next read
    */
   staleTime?: number
@@ -106,6 +112,7 @@ export const USE_QUERY_DEFAULTS = {
   refetchOnWindowFocus: true as _RefetchOnControl,
   refetchOnReconnect: true as _RefetchOnControl,
   refetchOnMount: true as _RefetchOnControl,
+  enabled: (() => true) as WatchSource<boolean>,
   // as any to simplify the typing with generics
   transformError: (error: unknown) => error as any,
 } satisfies Partial<UseQueryOptions>
--- a/src/query-options.ts
+++ b/src/query-options.ts
@@ -1,4 +1,4 @@
-import { type InjectionKey, type MaybeRefOrGetter, type WatchSource, inject } from 'vue'
+import { type InjectionKey, type MaybeRefOrGetter, inject } from 'vue'
 import type { EntryKey } from './entry-options'
 import type { QueryPluginOptions } from './query-plugin'
 import type { ErrorDefault } from './types-extension'
@@ -64,7 +64,7 @@ export interface UseQueryOptions<TResult = unknown, TError = ErrorDefault> {
    * Reactive boolean, indicating whether refetch/refresh should be called internally,
    * calling refetch when toggled to true
    */
-  enabled?: WatchSource<boolean>
+  enabled?: MaybeRefOrGetter<boolean>
 
   /**
    * Time in ms after which the data is considered stale and will be refreshed on next read
@@ -112,7 +112,7 @@ export const USE_QUERY_DEFAULTS = {
   refetchOnWindowFocus: true as _RefetchOnControl,
   refetchOnReconnect: true as _RefetchOnControl,
   refetchOnMount: true as _RefetchOnControl,
-  enabled: (() => true) as WatchSource<boolean>,
+  enabled: true as MaybeRefOrGetter<boolean>,
   // as any to simplify the typing with generics
   transformError: (error: unknown) => error as any,
 } satisfies Partial<UseQueryOptions>
--- a/src/use-query.spec.ts
+++ b/src/use-query.spec.ts
@@ -532,8 +532,8 @@ describe('useQuery', () => {
       expect(query).toHaveBeenCalledTimes(2)
     })
 
-    it('does not refresh/refetch when disabled', async () => {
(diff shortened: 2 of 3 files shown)


src/query-options.ts
```diff
@@ -60,6 +60,12 @@ export interface UseQueryOptions<TResult = unknown, TError = ErrorDefault> {
    */
   query: (context: UseQueryFnContext) => Promise<TResult>
 
+  /**
+   * Reactive boolean, indicating whether refetch/refresh should be called internally,
+   * calling refetch when toggled to true
+   */
+  enabled?: WatchSource<boolean>
```
posva: Let's use `MaybeRefOrGetter` instead

src/query-options.ts
```diff
@@ -106,6 +112,7 @@ export const USE_QUERY_DEFAULTS = {
   refetchOnWindowFocus: true as _RefetchOnControl,
   refetchOnReconnect: true as _RefetchOnControl,
   refetchOnMount: true as _RefetchOnControl,
+  enabled: (() => true) as WatchSource<boolean>,
```
posva:
```suggestion
  enabled: true,
```

src/use-query.spec.ts
```diff
@@ -532,6 +532,43 @@ describe('useQuery', () => {
       expect(query).toHaveBeenCalledTimes(2)
     })
 
+    it('does not refresh/refetch when disabled', async () => {
+      const { wrapper, query } = mountDynamicKey({ enabled: () => false })
```
posva: let's simplify the tests too by just using `false`

src/use-query.spec.ts
```diff
@@ -532,6 +532,43 @@ describe('useQuery', () => {
       expect(query).toHaveBeenCalledTimes(2)
     })
 
+    it('does not refresh/refetch when disabled', async () => {
```
posva:
```suggestion
    it('does not automatically refresh when enabled is false', async () => {
```

I think it's better to explicitly use the name of the option for code search puposes

src/use-query.spec.ts
```diff
@@ -532,6 +532,43 @@ describe('useQuery', () => {
...
+      // should refresh manually
+      await wrapper.vm.refresh()
+      await flushPromises()
+      expect(query).toBeCalledTimes(1)
+
+      await wrapper.vm.setId(2)
+      expect(query).toBeCalledTimes(1)
+    })
+
+    it('refetches when enabled toggled', async () => {
```
posva:
```suggestion
    it('triggers the query function when enabled becomes true', async () => {
```

src/use-query.ts
```diff
@@ -73,6 +73,15 @@ export function useQuery<TResult, TError = ErrorDefault>(
   const refresh = () => store.refresh(entry.value)
   const refetch = () => store.refetch(entry.value)
 
+  // if passed by client
+  if (_options.enabled !== undefined) {
+    watch(options.enabled, (newEnabled, oldEnabled) => {
+      // add test case why we check !oldEnabled
+      // add test case when we update entry, add test case when we update enabled
+      if (!oldEnabled && newEnabled) refetch()
```
posva:
```suggestion
      if (!oldEnabled && newEnabled) refresh()
```

It should always be refresh by default. Maybe it's worth adding another unit test that ensures the query function is not called when enabled goes from true to false to true within the staletime

src/use-query.ts
```diff
@@ -153,10 +162,12 @@ export function useQuery<TResult, TError = ErrorDefault>(
         // TODO: refactor noce we introduce the enabled option
         || queryReturn.status.value === 'pending'
       ) {
-        if (options.refetchOnMount === 'always') {
-          refetch()
-        } else {
-          refresh()
+        if (toValue(options.enabled)) {
```
posva: let's move it to the condition above `if (enabled && (...))`

src/use-query.ts
```diff
@@ -173,21 +184,25 @@ export function useQuery<TResult, TError = ErrorDefault>(
     if (options.refetchOnWindowFocus) {
       useEventListener(document, 'visibilitychange', () => {
         if (document.visibilityState === 'visible') {
-          if (options.refetchOnWindowFocus === 'always') {
-            refetch()
-          } else {
-            refresh()
+          if (toValue(options.enabled)) {
```
posva: let's move it to the condition above

posva (changes requested):
Nice! Thanks a lot

quiteeasy:
Thank you for taking the time to review my pull request. I appreciate the thoughtful feedback you provided. Your suggestions were well-reasoned, and I have implemented the recommended changes.

quiteeasy:
Should I resolve the threads you created, or would you prefer to do it yourself?

src/use-query.ts
```diff
@@ -73,6 +74,17 @@ export function useQuery<TResult, TError = ErrorDefault>(
   const refresh = () => store.refresh(entry.value)
   const refetch = () => store.refetch(entry.value)
 
+  // if passed by client
+  // & check if watchable
+  if (_options.enabled !== undefined
+      && (isRef(options.enabled) || typeof options.enabled === 'function')) {
```
posva: wouldn't checking the typeof to be boolean with options be enough? Since we have the defaults, it should be enough to skip adding a watcher for a primitive

src/use-query.ts
posva: just this one missing

posva:
Feel free to resolve them. I thought it would be automatic when committing

src/use-query.ts
quiteeasy: oh missed this one, thanks for pointing

src/use-query.ts
quiteeasy: thank you, looks much better! I have added another commit

src/use-query.ts
```diff
@@ -73,6 +73,16 @@ export function useQuery<TResult, TError = ErrorDefault>(
   const refresh = () => store.refresh(entry.value)
   const refetch = () => store.refetch(entry.value)
 
+  // if passed by client
+  // & check if watchable
+  if (_options.enabled !== undefined && typeof options.enabled !== 'boolean') {
```
posva: since options merges the default value of `true`, this should also work, right?

```suggestion
  if (typeof options.enabled !== 'boolean') {
```

posva (approved):
Thanks!

PR #46, opened 2024-05-20
feat: add mutation plugin

### LINKED ISSUE #15: MutationPlugin
Add a mutation plugin that allows setting up global options (onSuccess, etc that run after local ones)

quiteeasy:
Close #15 

This pull request introduces the initial implementation of the mutation plugin.

I noticed a couple of TODOs in the code (specifically in the useMutation hook, stating "TODO: move to plugin"). However, I am uncertain about moving the method responsible for invalidating cache entries to the plugin.

Relocating this functionality to the plugin would require passing mutation-scoped variables like `vars` and `newData`. This could potentially introduce coupling and complexity, as these variables are specific to individual mutations. Additionally, I'm uncertain about the tangible benefits of this approach.

Discarding cache entries within each respective mutation implementation might be more cohesive and aligned with the principle of separation of concerns.

I would appreciate your thoughts and feedback on this matter. Please share your perspective on the trade-offs and potential implications of moving the cache invalidation logic to the plugin versus keeping it within the mutation implementations.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,11 +50,14 @@ Install the plugins for the features you need:
 
 ```js
 import { createPinia } from 'pinia'
-import { QueryPlugin } from '@pinia/colada'
+import { MutationPlugin, QueryPlugin } from '@pinia/colada'
 
 app.use(createPinia())
 // install after pinia
 app.use(QueryPlugin, {
+  // optional options
+})
+app.use(MutationPlugin, {
   // optional options
 })
 ```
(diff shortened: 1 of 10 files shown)


posva:
Thanks! The todo was referring to another kind of plugin, something not yet in the codebase. Some todos are not explicit enough 😅
I appreciate the effort thought ! I will try to rework the PR to give you a review

posva:
Thanks for this! In the end I decided to make it easy to consume mutations and allow mutations options to be passed directly to PiniaColada instead of having two separate plugins.

PR #61, opened 2024-07-09
chore(playground): fix the playground auto routers error

zhangmo8:
When I try to launch the playground in the project. It reported an error because routes were not import from auto-routes


posva (approved):
Thanks! The playground libs auto updated but it had breaking changes and wasn't tested

PR #65, opened 2024-08-04
feat(mutation): refetch active queries

ElisePatrikainen:


ElisePatrikainen:
@posva I think we could also add a mutation status for the time when the active invalidated queries are being refetch. That could be useful in some cases of optimistic updates, for example in the case described here : https://github.com/posva/pinia-colada/pull/67/files#diff-5b006d752fa7c238bea51c764a51ae11a15731bc832d2ada6796dce589ffa361R174

posva:
I was thinking of keeping the state split as well to make it similar to useQuery(). Instead of having `queryStatus`, have `mutationStatus` + the same `status` as query. Ideally, it would be named the same but I can't find a good name, `opStatus`, `operationStatus`, `asyncStatus`, `taskStatus`?

ElisePatrikainen:
This sounds great. I would say `asyncStatus` is the coolest, and `operationStatus` and `taskStatus` the clearest. `taskStatus` has the advantage of being shorter than `operationStatus` (`opStatus` being guessable though). Therefore my preference would go for `taskStatus`.

ElisePatrikainen:
For the refresh of invalidated queries, shall it be then included in the `pending` status? 

The thing is that, currently, it is difficult to use the mutation's `pending`/`success` statuses on their own (because, when the status changes, the mutation is not 'visible' yet since the refresh is not done). I am thinking that this might affect more use cases than only optimistic updates (displaying a loader during the mutation for example), even though most of use cases can probably be handled by combining the mutation and invalidated queries states (for displaying a loader during the mutation: `mutationPending || invalidatedQueryPending`). On the other hand, including the auto-refetch in the `pending` status implies that, in case of failure of a refetch, the mutation is considered as failed, even though it is not, which might lead to inconstancies.

To summarise, pros and cons of including the auto-refetch in the `pending` status:
- pros: the mutation `pending` (and `success`) status can be used on its own (no need to combine it with the refetch queries status), enables more specific use cases (having different loaders for the queries and for the mutation -which would be displayed during the the mutation and the auto-refetch-, optimistic updates...)
- cons: ties the mutation state to the refetch queries state (which can lead to inconsistencies in case of failure of a refetch)

src/query-store.ts
```diff
@@ -333,9 +333,17 @@ export const useQueryCache = defineStore(QUERY_STORE_ID, ({ action }) => {
    * @param options.refetch - if true, it will refetch the data
    */
   const invalidate = action((entry: UseQueryEntry) => {
-    // will force a fetch next time
-    entry.when = 0
+    // TODO: once `fetch` cancels query, move `cancelQuery` to the second condition
     cancelQuery(entry)
+    if (entry.deps.size > 0) {
+      // reset any pending request
+      entry.pending = null
```
posva: this should mamybe kept as the pending request is no longer valid

src/query-store.ts
```diff
@@ -333,9 +333,17 @@ export const useQueryCache = defineStore(QUERY_STORE_ID, ({ action }) => {
...
+      // force refresh
+      fetch(entry)
```
posva: I don't think the invalidate should fetch. I think it's better to add a different action that fetches only active queries or maybe introduce a readonly property on the entry `active` or similar that can be checked on the mutation side

ElisePatrikainen: That's a good point. Since I haven't found other use cases for it (checking if a query is active and refetching it), I moved it on the mutation side.

src/use-mutation.ts
```diff
@@ -255,6 +255,12 @@ export function useMutation<
           )) {
             // TODO: find a way to pass a source of the invalidation, could be a symbol associated with the mutation, the parameters
             store.invalidate(entry)
+            // auto refresh of the active queries
+            if (entry.active) {
+              // reset any pending request
+              entry.pending = null
```
posva: We can remove this since it's done by invalidate when cancelling the query
```suggestion
```

ElisePatrikainen: Indeed, done.

posva (approved):

PR #78, opened 2024-09-18
feat(query): placeholder

ElisePatrikainen:


src/use-query.ts
```diff
@@ -200,6 +200,15 @@ export function useQuery<TResult, TError = ErrorDefault>(
     entry,
     (entry, previousEntry) => {
       if (!isActive) return
+      if (entry.stale && entry.options?.placeholderData && previousEntry?.state.value.status === 'success') {
+        cacheEntries.setEntryState(entry, {
+          // TODO: pending?
```
posva: the placeholder data cannot change the cache. It should only appear as a fallback while the entry is not in success state

ElisePatrikainen: Understood, this makes sens. I changed the approach regarding it (to store the placeholder data in the entry cache but not in the entry state, as I had done first), let me know what you think about it.

The placeholder fallback should also appear in error state, right?

posva: I think so. The point is to have fallback data while it’s undefined

ElisePatrikainen: Perfect.

posva:
I pushed to feat/placeholder-data. I think it will be a better starting point. Could you port the tests to that branch, push there and open a PR with it?

ElisePatrikainen:
> I pushed to feat/placeholder-data. I think it will be a better starting point. Could you port the tests to that branch, push there and open a PR with it?

Sure, here it is: https://github.com/posva/pinia-colada/pull/80

PR #80, opened 2024-09-22
feat(query): placeholderData

### LINKED ISSUE #57: Add a placeholder data option
Loading spinners are bad, it less disruptive to show users previous data until new data is ready. Suspense is somewhat based on that idea.

Have code like this:
```ts
const { data } = useQuery({
	key: () => ['item', props.selectedItemId],
	query: () => getItemById(props.selectedItemId),
})
```
when props.selectedItemId changes current behavior of useQuery forces me to show blank screen because data does not exist, I wish data would not be reset while isFetching is true.

ElisePatrikainen:
Closes #57

### DIFF
--- a/src/query-store.ts
+++ b/src/query-store.ts
@@ -41,7 +41,7 @@ export interface UseQueryEntry<TResult = unknown, TError = unknown> {
    * A placeholder `data` that is initially shown while the query is loading for the first time. This will also show the
    * `status` as `success` until the query finishes loading (no matter the outcome).
    */
-  placeholderData: TResult | null
+  placeholderData: TResult | null | undefined | void
 
   /**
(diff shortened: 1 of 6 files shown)


src/query-store.ts
```diff
@@ -37,6 +37,11 @@ export interface UseQueryEntry<TResult = unknown, TError = unknown> {
    */
   state: ShallowRef<DataState<TResult, TError>>
 
+  /**
+   * A placeholder data that is initially shown while the query is loading for the first time.
```
ElisePatrikainen: @posva: should the error case be mentioned?

posva: yes but in the options definition, not here

PR #97, opened 2024-11-07
feat(plugins): add auto-refetch plugin

### LINKED ISSUE #49: Add `autoRefetch` Option to `useQuery` for automatic stale entry refetching
Add a boolean option, `autoRefetch`, to automatically refetch an entry when it becomes stale.

ymansurozer:
Closes #49.

This PR adds a new plugin that enables automatic query refetching when data becomes stale. This provides a simple way to keep data fresh without manual intervention.

**Features**
- Automatically refreshes queries when their staleTime is reached
- Can be enabled/disabled globally via plugin options
- Can be controlled per-query via autoRefetch option
- Cleans up properly when queries are unmounted
- Only runs on the client side

**Issues and Questions**
- I am having an issue with one of the tests. To be honest, I am not good with writing tests (this is one of the rare cases where I've written one). So some help would be great!
- The refetch timeout is currently set in `refresh` hook. I also tried `fetch` but the issue was `fetch` is not triggered on initial page load but `refresh` does. So even if the user does nothing regarding the query from initial load, it will be auto-fetched in the background.
- To trigger a refresh, I've used `queryCache.refresh(entry)`. I noticed there are also options like `fetch`, `invalidate`, and `invalidateQueries`. I used `refresh` because I understand it has an internal check on the data to see if it is fresh. I did not want to force it with `fetch`.
- The refetch timeouts are only set on the client. I think this should be the expected behavior, as otherwise the refetch keeps happening on the server, too.

I would love your feedback on this. I am happy to work on this further.

### DIFF
--- a/plugins/auto-refetch/README.md
+++ b/plugins/auto-refetch/README.md
@@ -24,7 +24,7 @@ import { PiniaColadaAutoRefetch } from '@pinia/colada-plugin-auto-refetch'
 app.use(PiniaColada, {
   // ...
   plugins: [
-    PiniaColadaAutoRefetch(),
+    PiniaColadaAutoRefetch({ autoRefetch: true }), // enable globally
   ],
 })
 ```
--- a/plugins/auto-refetch/README.md
+++ b/plugins/auto-refetch/README.md
@@ -35,7 +35,7 @@ You can customize the refetch behavior individually for each query with the `aut
 useQuery({
   key: ['todos'],
   query: getTodos,
-  autoRefetch: false, // disable auto refetch
+  autoRefetch: true, // override local autoRefetch
 })
 ```
 
--- a/plugins/auto-refetch/README.md
+++ b/plugins/auto-refetch/README.md
@@ -1,0 +1,44 @@
+<h1>
+  <img height="76" src="https://github.com/posva/pinia-colada/assets/664177/02011637-f94d-4a35-854a-02f7aed86a3c" alt="Pinia Colada logo">
+  Pinia Colada Auto Refetch
+</h1>
+
+<a href="https://npmjs.com/package/@pinia/colada-plugin-auto-refetch">
+  <img src="https://badgen.net/npm/v/@pinia/colada-plugin-auto-refetch/latest" alt="npm package">
+</a>
+
+Automatically refetch queries when they become stale in Pinia Colada.
+
+## Installation
+
+```sh
+npm install @pinia/colada-plugin-auto-refetch
+```
+
+## Usage
+
+```js
+import { PiniaColadaAutoRefetch } from '@pinia/colada-plugin-auto-refetch'
+
+// Pass the plugin to Pinia Colada options
+app.use(PiniaColada, {
+  // ...
+  plugins: [
+    PiniaColadaAutoRefetch(),
+  ],
+})
+```
+
+You can customize the refetch behavior individually for each query with the `autoRefetch` option:
+
+```ts
+useQuery({
+  key: ['todos'],
+  query: getTodos,
+  autoRefetch: false, // disable auto refetch
+})
+```
+
+## License
+
+[MIT](http://opensource.org/licenses/MIT)
--- a/plugins/auto-refetch/package.json
+++ b/plugins/auto-refetch/package.json
@@ -60,7 +60,7 @@
   "scripts": {
     "build": "tsup",
     "changelog": "conventional-changelog -p angular -i CHANGELOG.md -s --commit-path . -l @pinia/colada-plugin-auto-refetch -r 1",
-    "test": "echo \"Error: no test specified\" && exit 1"
+    "test": "vitest --ui"
   },
   "peerDependencies": {
     "@pinia/colada": "workspace:^"
--- a/plugins/auto-refetch/package.json
+++ b/plugins/auto-refetch/package.json
@@ -1,0 +1,71 @@
+{
+  "name": "@pinia/colada-plugin-auto-refetch",
+  "type": "module",
+  "publishConfig": {
+    "access": "public"
+  },
+  "version": "0.0.1",
+  "description": "Automatically refetch queries when they become stale in Pinia Colada",
+  "author": {
+    "name": "Yusuf Mansur Ozer",
+    "email": "ymansurozer@gmail.com"
+  },
+  "license": "MIT",
+  "homepage": "https://github.com/posva/pinia-colada/plugins/auto-refetch#readme",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/posva/pinia-colada.git"
+  },
+  "bugs": {
+    "url": "https://github.com/posva/pinia-colada/issues"
+  },
+  "keywords": [
+    "pinia",
+    "plugin",
+    "data",
+    "fetching",
+    "query",
+    "mutation",
+    "cache",
+    "layer",
+    "refetch"
+  ],
+  "sideEffects": false,
+  "exports": {
+    ".": {
+      "types": {
+        "import": "./dist/index.d.ts",
+        "require": "./dist/index.d.cts"
+      },
+      "import": "./dist/index.js",
+      "require": "./dist/index.cjs"
+    }
+  },
+  "main": "./dist/index.cjs",
+  "module": "./dist/index.js",
+  "types": "./dist/index.d.ts",
+  "typesVersions": {
+    "*": {
+      "*": [
+        "./dist/*",
+        "./*"
+      ]
+    }
+  },
+  "files": [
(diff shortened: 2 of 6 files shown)


plugins/auto-refetch/package.json
```diff
@@ -0,0 +1,71 @@
...
+  },
+  "files": [
+    "LICENSE",
+    "README.md",
+    "dist"
+  ],
+  "scripts": {
+    "build": "tsup",
+    "changelog": "conventional-changelog -p angular -i CHANGELOG.md -s --commit-path . -l @pinia/colada-plugin-auto-refetch -r 1",
+    "test": "echo \"Error: no test specified\" && exit 1"
```
posva:
```suggestion
    "test": "vitest --ui"
```

plugins/auto-refetch/src/index.ts
```diff
@@ -0,0 +1,78 @@
...
+  }
+}
+
+// Add types for the new option
+declare module '@pinia/colada' {
+  interface UseQueryOptions {
+    /**
+     * Whether to automatically refresh this query when it becomes stale.
+     */
+    autoRefetch?: boolean
```
posva: I think it's worth supporting a `MaybeRefOrGetter<boolean>`

plugins/auto-refetch/src/index.ts
```diff
@@ -0,0 +1,78 @@
...
+  return ({ queryCache }) => {
+    // Keep track of active entries and their timeouts
+    const refetchTimeouts = new Map<string, NodeJS.Timeout>()
+
+    queryCache.$onAction(({ name, args, after }) => {
+      // We want refetch to happen only on the client
+      if (!import.meta.client) return
+
+      // Handle fetch to set up auto-refetch
+      if (name === 'refresh') {
```
posva: `refresh()` is indeed called more often than `fetch()`. Maybe we need a combination of `fetch` and `ensure` to make it work at creation too

plugins/auto-refetch/src/index.ts
```diff
@@ -0,0 +1,78 @@
...
+  enabled?: boolean
+}
+
+/**
+ * Plugin that automatically refreshes queries when they become stale
+ */
+export function PiniaColadaAutoRefetch(
+  options: PiniaColadaAutoRefetchOptions = {},
+): PiniaColadaPlugin {
+  const { enabled = true } = options
```
posva: default to `false` is probably better, no?

posva (changes requested):
Nice! I will come back for tests later

plugins/auto-refetch/src/index.ts
ymansurozer: Sure!

plugins/auto-refetch/src/index.ts
ymansurozer: My initial thinking was if someone is adding this plugin their intention would be to enable auto fetch. But I now think false is better considering it might be a large project and the plugin might be added to use it on some queries. Changing it now.

plugins/auto-refetch/src/index.ts
ymansurozer: Makes sense. I refactored to ensure + fetch now. But I had to remove the `entry.active` check because `ensure` hook returns `useQueryOptions` instead of `UseQueryEntry`, but I am guessing that should be fine.

ymansurozer:
Thanks for the feedback @posva! Sent some commits to refactor based on your comments.

plugins/auto-refetch/src/auto-refetch.spec.ts
```diff
@@ -0,0 +1,129 @@
...
+            ...queryOptions,
+          })
+        },
+      }),
+      {
+        global: {
+          plugins: [
+            createPinia(),
+            [PiniaColada, {
+              plugins: [PiniaColadaAutoRefetch(pluginOptions)],
```
posva: you probably need to pass `enabled: true` by default

plugins/auto-refetch/src/auto-refetch.spec.ts
```diff
@@ -0,0 +1,129 @@
...
+    expect(query).toHaveBeenCalledTimes(1)
+
+    // Advance time past stale time in one go
+    vi.advanceTimersByTime(1000)
+    await flushPromises()
+
+    expect(query).toHaveBeenCalledTimes(2)
+  })
+
+  it('respects disabled option globally', async () => {
```
posva: Prefer using enabled, it's easier to search since the option is named enabled
```suggestion
  it('respects enabled option globally', async () => {
```

plugins/auto-refetch/src/auto-refetch.spec.ts
```diff
@@ -0,0 +1,129 @@
...
+
+    await flushPromises()
+    expect(query).toHaveBeenCalledTimes(1)
+
+    vi.advanceTimersByTime(2000)
+    await flushPromises()
+    expect(query).toHaveBeenCalledTimes(1)
+  })
+
+  it('cleans up timeouts when query is unmounted', async () => {
```
posva:
```suggestion
  it('avoids refetching an unactive query', async () => {
```

plugins/auto-refetch/src/index.ts
```diff
@@ -0,0 +1,107 @@
+import type { PiniaColadaPlugin, UseQueryOptions } from '@pinia/colada'
+import type { MaybeRefOrGetter } from 'vue'
+import { toValue } from 'vue'
+
+export interface PiniaColadaAutoRefetchOptions {
+  /**
+   * Whether to enable auto refresh by default.
+   * @default false
+   */
+  enabled?: boolean
```
posva: let's rename it to match the per-query option
```suggestion
  autoRefetch?: boolean
```

posva (changes requested):
There should also be another test that ensures the timer is reset if a new request is fired before the data becomes stale

ymansurozer:
> There should also be another test that ensures the timer is reset if a new request is fired before the data becomes stale

Thank you @posva! I've added a new test for this and also accepted & implemented your suggestions. However, two tests are failing:
- `it automatically refetches when stale time is reached`
- `it resets the stale timer when a new request occurs`

In both cases, it seems the query was called one less time than expected. I tried to set stale time to 1ms or advancing time by 10000ms but the number of queries remain the same, so I think something in my tests are wrong but I cannot figure out what exactly. Tried using `fetch` instead of `refresh` in the plugin but that also did not work. Would you have any ideas?

plugins/auto-refetch/README.md
```diff
@@ -0,0 +1,44 @@
...
+})
+```
+
+You can customize the refetch behavior individually for each query with the `autoRefetch` option:
+
+```ts
+useQuery({
+  key: ['todos'],
+  query: getTodos,
+  autoRefetch: false, // disable auto refetch
```
posva:
```suggestion
  autoRefetch: true, // override local autoRefetch
```

plugins/auto-refetch/README.md
```diff
@@ -0,0 +1,44 @@
...
+## Usage
+
+```js
+import { PiniaColadaAutoRefetch } from '@pinia/colada-plugin-auto-refetch'
+
+// Pass the plugin to Pinia Colada options
+app.use(PiniaColada, {
+  // ...
+  plugins: [
+    PiniaColadaAutoRefetch(),
```
posva:
```suggestion
    PiniaColadaAutoRefetch({ autoRefetch: true }), // enable globally
```

plugins/auto-refetch/src/index.ts
```diff
@@ -0,0 +1,109 @@
...
+): PiniaColadaPlugin {
+  const { autoRefetch = false } = options
+
+  return ({ queryCache }) => {
+    // Keep track of active entries and their timeouts
+    const refetchTimeouts = new Map<string, NodeJS.Timeout>()
+
+    queryCache.$onAction(({ name, args, after }) => {
+      // We want refetch to happen only on the client
+      if (typeof document === 'undefined') return
```
posva: we can move this outside and do nothing if it's on server

ymansurozer:
All done! The only issue left is the `it resets the stale timer when a new request occurs` test, which I am not able to pass despite clearing timeouts before scheduling the refetch and at the time of the refetch.

posva (approved):
Thanks a ton for the great contribution! It turns out your test was good! Calling the `query` mock directly wasn't triggering things, but calling the `refetch` method made it work with no code changes!
I also removed the need for a map by storing the timeout in the entry itself, simplifying things and also fixing possible bugs with the serialization of the key (BTW I exposed now a `toCacheKey` function that could be used)

PR #98, opened 2024-11-09
feat(query): staleTime of Infinity always fresh.

nakagam3:
This PR modifies the `staleTime` option to accept `Infinity` as a valid value. If `Infinity` is specified, the cache will remain fresh indefinitely after the initial fetch, unless manually refetched.

This is useful in cases where the data doesn't update frequently, so refetching is unnecessary, and we may want to disable it for performance reasons.

I welcome your feedback on this change and am happy to make any necessary adjustments.

### DIFF
--- a/scripts/release.mjs
+++ b/scripts/release.mjs
@@ -1,4 +1,3 @@
-/* eslint-disable no-console */
 import fs from 'node:fs/promises'
 import { existsSync } from 'node:fs'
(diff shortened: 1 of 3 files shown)


posva (commented):
Thanks I pushed an alternative way to #99 because I couldn't push to your PR

nakagam3:
Thank you for your kindness! Your approach was really nice.

Issue #6, opened 2024-02-19
Preload queries


posva:
Allowing a query to be preloaded
Requires all the options rather than just the key as otherwise it might not exist

quiteeasy:
Is preloading a query the same as instantiating the query before user navigation? The key difference should be allowing preloading in user actions (e.g., click, hover, etc.) and in lifecycle hooks. Should we introduce another API, like `usePreloadQuery`, specifically for this purpose?

posva:
it is about instantiating the query. It doesn't have to be about navigation but it should integrate easily into that too. I'm not sure if a new API is needed yet. The idea is to have:

- A way to ensure the entry (should be good)
- A method that can load the entry and return a promise

So maybe this is just a matter of documentation at this point. Not sure yet

posva:
Currently doable with

```ts
const queryCache = useQueryCache()
queryCache.setQueryData(['the key'], await fetchData())
```

Depending on the `staleTime`, the data will be fresh and calling `refresh()` won't get it again

OsamaHaikal:
Is this the same as we do await useAsyncData with immediate true to await the results?

posva:
> Is this the same as we do await useAsyncData (in nuxt) with immediate true to await the results?

no

posva:
https://pinia-colada.esm.dev/cookbook/prefetching.html

PR #103, opened 2024-11-12
feat: prefetch on queries

brenoliradev:
Hey. This is a draft PR heavily inspired on: https://github.com/TanStack/query/blob/f56fc8415904d79a4658167e01a8dbbcf842bd8f/packages/query-core/src/queryClient.ts#L357

Aims to solve: https://github.com/posva/pinia-colada/issues/6

I tried to mimic the behavior but using `queryCache`.

Ps: The idea behind this PR is to start a discussion about if it's viable to prefetch in this way or not.


brenoliradev:
@posva Hey! I think it's good to go for a first review.

src/query-store.ts
```diff
@@ -570,6 +570,25 @@ export const useQueryCache = /* @__PURE__ */ defineStore(QUERY_STORE_ID, ({ acti
...
+      const options: UseQueryOptionsWithDefaults<TResult, TError> = {
+        ...optionDefaults,
+        ...opts,
+      }
+
+      const entry = ensure<TResult, TError>(options)
+
+      refresh(entry).then(() => undefined).catch(() => undefined)
+    },
+  )
```
posva: Let's:

- make the function a one liner
- remove `action` because this won't be called within the store
- return the refresh call (implicit in a one-liner)
- replace the two `() => undefined` with the noop used in `use-query` (extract to src/utils if needed)

src/use-query.spec.ts
```diff
@@ -1208,4 +1208,77 @@ describe('useQuery', () => {
...
+
+      await flushPromises()
+
+      expect(wrapper.vm.data).toBe(42)
+
+      expect(query).toHaveBeenCalledOnce()
+
+      const queryCache = useQueryCache()
+
+      queryCache.prefetch({
```
posva, line 1222: We should be able to await prefetch instead of flushing promises

src/use-query.spec.ts
```diff
@@ -1208,4 +1208,77 @@ describe('useQuery', () => {
       // No warnings!
     })
   })
+
+  describe('prefetchs query', () => {
+    it('should not prefetch a valid query', async () => {
```
posva:
```suggestion
    it('should not prefetch a fresh query', async () => {
```

src/use-query.spec.ts
```diff
@@ -1208,4 +1208,77 @@ describe('useQuery', () => {
...
+
+      expect(wrapper.vm.data).toBeUndefined()
+
+      await flushPromises()
+
+      expect(wrapper.vm.data).toBe(42)
+
+      expect(query).toHaveBeenCalledOnce()
+
+      const queryCache = useQueryCache()
```
posva: Let's reduce the test surface:

```suggestion
      await flushPromises()

      const queryCache = useQueryCache()
```

src/use-query.spec.ts
```diff
@@ -1208,4 +1208,77 @@ describe('useQuery', () => {
...
+      queryCache.prefetch({
+        key: ['key'],
+        query,
+      })
+
+      await flushPromises()
+
+      expect(wrapper.vm.data).toBe(42)
+
+      expect(query).toHaveBeenCalledOnce()
```
posva: Let's reduce a bit the empty lines
```suggestion
      expect(wrapper.vm.data).toBe(42)
      expect(query).toHaveBeenCalledOnce()
```

posva (changes requested):
Thanks, it's looking good!
I still need to look at this from the practical perspective of using it with data loaders before going further so it might take a few weeks to merge

brenoliradev:
@posva I think everything is fixed!

posva:
I feel like documenting how to prefetch with

```ts
const queryCache = useQueryCache()
queryCache.setQueryData(['the key'], await fetchData())
```

is enough. When thinking about the API, you always need a valid key (no dynamic values), so the only case that could benefit from it is a static defineQuery `defineQuery({})` (object instead of fn). What do you think? The main issue is that prefetching a query that contains a dynamic value (like a route params) does not work because when prefetching, we want to prefetch a different value, so having full control is better

https://github.com/posva/pinia-colada/issues/6#issuecomment-2457628444

brenoliradev:
I think in a long term having a separated prefetch would to implement future plugins.

brenoliradev:
It could be. 

I think the greater difference here from `prefetchQuery` (from react query) is the support for more complicated behavior (for example, the `select` that we discussed before). I imagine that having a dedicated prefetch API would simplify supporting different plugins.

I'm not entirely sure, but I by using the behavior you provided I wouldn't have control over query options.

In other way, keeping it minimal seems to be one of core concepts in `pinia-colada`. Personally, I would prefer having more fine-grained control.

The response for this may reside in the creation of a dedicated prefetch plugin if the user want more fine-grained support. What do you think?

posva:
That’s the thing: options do not interfere in prefetching, that’s why I think it’s worth keeping it simple and direct.

regarding select, I don’t feel like it add any value compared to just writing the code within the query function

brenoliradev:
I thought my implementation was covering options. Sorry. Why isn't it tho? I remember to include an options handler when defining the query. 

Yeah! I remember that. That's why I imagine: doing it straightfoward is the `pinia-colada` way of doing it.

posva:
The implementation does cover options as it passes them but since we are just doing the initial fetch to populate the data and avoid another fetch (that’s what I consider the goal of prefetch), only the key and the data are used

brenoliradev:
now I got it. we can go for the path of adding options (I can drop a PR fully covering them if any of them is working) or just do as you want to add to docs as a minimal way to prefetch it

posva:
I think adding the minimal example to docs is the safest path for the moment

posva:
Thanks for this. I added it to docs

Issue #111, opened 2024-11-19
`select` option for queries


brenoliradev:
Hey! Just a feature request for: https://tanstack.com/query/latest/docs/framework/react/guides/render-optimizations#select

I'm on my way to building features to use `pinia-colada` on my project on the future instead of `vue-query` so I'll work in this. Just creating the issue so we can discuss it here too.

posva:
I'm actually not a big fan of `select`, I don't think it's needed for Vue because the 

> This is useful for highly optimized data transformations or to avoid unnecessary re-renders.

doesn't apply.

I would personally just write that logic within the `query` function. I'm open to changing my mind of course.

PR #112, opened 2024-11-19
Feat/select

brenoliradev:
PR for: https://github.com/posva/pinia-colada/issues/111

### DIFF
--- a/src/query-options.ts
+++ b/src/query-options.ts
@@ -155,6 +155,12 @@ export interface UseQueryOptions<TResult = unknown, TError = ErrorDefault> {
    * @default true
    */
   refetchOnReconnect?: MaybeRefOrGetter<_RefetchOnControl>
+
+  /**
+   * Transforms the query data to `TData`
+   * TODO: write better doc
+   */
+  select?: (data: NoInfer<TResult>) => NoInfer<TResult>
 }
 
 /**
--- a/src/query-store.ts
+++ b/src/query-store.ts
@@ -603,23 +603,29 @@ export const useQueryCache = /* @__PURE__ */ defineStore(QUERY_STORE_ID, ({ acti
       state: DataState<NoInfer<TResult>, NoInfer<TError>>,
     ) => {
       entry.when = Date.now()
-
-      if (typeof entry.options?.select === 'function' && state.status === 'success') {
-        try {
-          entry.state.value = { ...state, data: entry.options.select(state.data) }
-        } catch (e) {
-          // TODO: check how to handle the error state
-          entry.state.value = {
-            status: 'error',
-            data: undefined,
-            error: e as TError,
-          }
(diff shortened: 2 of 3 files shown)


src/query-options.ts
```diff
@@ -155,6 +155,12 @@ export interface UseQueryOptions<TResult = unknown, TError = ErrorDefault> {
    * @default true
    */
   refetchOnReconnect?: MaybeRefOrGetter<_RefetchOnControl>
+
+  /**
+   * Transforms the query data to `TData`
+   * TODO: write better doc
+   */
+  select?: (data: NoInfer<TResult>) => NoInfer<TResult>
```
brenoliradev, line 163: Need help here. Should data be this or we do it as `any` (tanstack query does the latter)? 

Which type would work the best?

src/query-store.ts
```diff
@@ -839,3 +839,23 @@ export function hydrateQueryCache(
 export function serializeQueryCache(queryCache: QueryCache): UseQueryEntryNodeSerialized[] {
   return serializeTreeMap(queryCache.caches)
 }
+
+/**
+ * Initial handler for select
+ */
+const parseSelectedData = <TResult, TError>(select: UseQueryOptions['select'],
```
brenoliradev, line 846: Is it ok to extract and reuse it around? 

It prob need a better jsdoc though

src/query-store.ts
```diff
@@ -839,3 +839,23 @@ export function hydrateQueryCache(
...
+  // NOTE: NoInfer ensures correct inference of TResult and TError
+  state: DataState<NoInfer<TResult>, NoInfer<TError>>): DataState<NoInfer<TResult>, NoInfer<TError>> => {
+    if (typeof select !== 'function' || state.status !== 'success') return state
+
+    try {
+      return { ...state, data: select(state.data) }
+    } catch (e) {
+      // TODO: check how to handle the error state
```
brenoliradev, line 854: Need help here too. Which kinda of error should be written there?

src/use-query.spec.ts
```diff
@@ -1208,4 +1208,60 @@ describe('useQuery', () => {
...
+        throw new Error('Invalid select function')
+      },
+    })
+
+    await flushPromises()
+
+    expect(wrapper.vm.state.error).toBeInstanceOf(Error)
+  })
+
+  // TODO discuss if it should or not
```
brenoliradev, line 1255: Should it updates `initialData`?

brenoliradev (commented):
hey @posva when possible give me pointers here! 

best,

posva:
See https://github.com/posva/pinia-colada/discussions/113

posva:
If you want to go forward with this, I think creating plugins for commonly used options would be the way! 👍

PR #128, opened 2024-12-07
Multi mutation hook

tsiotska:


posva (commented):
Thanks for the PR! I think this should be doable without touching much the mutation store but it might also be missing a few things yet.
If you want, you can give it a try without adding new actions to the mutation store

tsiotska:
I want, once I ensure base multi mutation works I plan to take a deeper look into the missing essential Mutation features, including:

- gcTime (garbage collection for stale cache)
- Requests batching
- Persisting cache (e.g., local storage)
- Automatic request retry
- Global loading indicator (e.g., isGlobalMutationLoading)
- networkMode
- Cancel hook etc.

posva:
I'm happy to have more contributions! Most of these are planned as plugins to keep the library to the most used features and keep it lean. `gcTime` is already implemented. Network mode I'm not sure about its utility. It should be doable as a plugin with a couple of small additions to the stores but I'm not even sure it's worth it. Do you have use cases where this feature proved to be critical?

tsiotska:
TanStack has abstraction with options for both query and mutation. I might be quite misled, but we have gcTime as part of UseQueryOptions interface and UseMutationOptions lacks some properties like gcTime. (We should also pay attention that queryClient is missing and maybe it would make lot's of reconstruction if being postponed).

I want to ask If I could use /playground/pages/contacts/new-usage-two-[id] for writing multiMutation playground test.

PR #146, opened 2024-12-27
bug/should-refetch-on-each-mount

jotolo:
#### Description

This PR adds a test for the `defineQuery` method.
It shows the behaviour of the query function when mounting/unmounting a component multiple times. 
As result we can see the test fails and the query method is only called on the first `mount`.

#### Expected behaviour

It should refetch the data with each onMounted

PD: Further investigation made me noticed it would refetch again on mount only if the key is not active in the cache anymore. 

@posva Thanks for creating this library. It is changing the way we do data-fetching in our team 🎉

### DIFF
--- a/src/define-query.spec.ts
+++ b/src/define-query.spec.ts
@@ -143,6 +143,40 @@ describe('defineQuery', () => {
       expect(spy).toHaveBeenCalledTimes(2)
     })
 
+    it('triggers as many times as it is mounted', async () => {
+      const spy = vi.fn(async () => 'ko')
+      const useDefinedQuery = defineQuery({
+        key: ['id'],
+        query: spy,
+      })
+
+      const pinia = createPinia()
+
+      const Comp1 = defineComponent({
+        setup() {
+          useDefinedQuery()
+          return {}
+        },
+        template: `<div></div>`,
+      })
+
+      const mountedComponent = mount(Comp1, {
+        global: {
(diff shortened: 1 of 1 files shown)


jotolo:
I just noticed while reading the code that if you set `refetchOnMount: 'always'` it does exactly what I was expecting. 
Is this expected behaviour? If so, can we include this particular part on the docs? 
https://pinia-colada.esm.dev/api/@pinia/colada/interfaces/UseQueryOptions.html#refetchOnMount-

I can open a PR to add this information: https://github.com/posva/pinia-colada/blob/b4d13b304d9bb9edab12c83e4ecddc1c359c03c3/src/query-options.ts#L6

posva:
I'm glad this is helping!

Yes, this behavior is intended: it refreshes by default. I think that what you want is to lower the staleTime of your query rather than using `refetchOnMount` always but not sure.

The information you are linking is already in the API docs

jotolo:
@posva I can see the default value is mentioned in the docs but not all possible values for `refetchOnMount`? I think all possible values should be included in the docs to avoid what happened to me 😅 Anyways, let me know if I can help with that.
Thanks for the quick reply 🤝

posva:
That's true! I made the `RefetchOnControl` type public to make it visible in the docs. Thanks!

jotolo:
Awesome @posva , that works too! Thanks

PR #153, opened 2025-01-07
feat(plugins): Add refetch interval plugin

vad1ym:
Hi! Colada is awesome! I’m planning to migrate from TanStack, but the key feature I’m currently missing to start using the library is `refetchInterval`.

I’ve added a base plugin and some test cases, following the structure of other plugins. It works and behaves similarly to TanStack, but Colada’s plugin API feels a bit complicated to me, so I’d appreciate your help in finishing it properly. 

In particular, I’m unsure about the Map keys I’m using. Should we use the query key (array?) itself as the Map key? And should it be a WeakMap instead?

### DIFF
--- a/plugins/refetch-interval/package.json
+++ b/plugins/refetch-interval/package.json
@@ -1,0 +1,73 @@
+{
+  "name": "@pinia/colada-plugin-refetch-interval",
+  "type": "module",
+  "publishConfig": {
+    "access": "public"
+  },
+  "version": "0.0.1",
+  "description": "Refetch queries on a regular interval with Pinia Colada",
+  "author": {
+    "name": "Eduardo San Martin Morote",
+    "email": "posva13@gmail.com"
+  },
+  "license": "MIT",
+  "funding": "https://github.com/sponsors/posva",
+  "homepage": "https://github.com/posva/pinia-colada/plugins/refetch-interval#readme",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/posva/pinia-colada.git"
+  },
+  "bugs": {
+    "url": "https://github.com/posva/pinia-colada/issues"
+  },
+  "keywords": [
+    "pinia",
+    "plugin",
+    "data",
+    "fetching",
+    "interval",
+    "refetch",
+    "query",
+    "mutation",
+    "cache",
+    "layer"
+  ],
+  "sideEffects": false,
+  "exports": {
+    ".": {
+      "types": {
+        "import": "./dist/index.d.ts",
+        "require": "./dist/index.d.cts"
+      },
+      "import": "./dist/index.js",
+      "require": "./dist/index.cjs"
+    }
+  },
+  "main": "./dist/index.cjs",
+  "module": "./dist/index.js",
+  "types": "./dist/index.d.ts",
+  "typesVersions": {
+    "*": {
+      "*": [
+        "./dist/*",
+        "./*"
+      ]
+    }
+  },
+  "files": [
+    "LICENSE",
+    "README.md",
+    "dist"
+  ],
+  "scripts": {
+    "build": "tsup",
+    "changelog": "conventional-changelog -p angular -i CHANGELOG.md -s --commit-path . -l @pinia/colada-plugin-refetch-interval -r 1",
+    "test": "vitest --ui"
+  },
+  "peerDependencies": {
+    "@pinia/colada": "workspace:^"
+  },
+  "devDependencies": {
+    "@pinia/colada": "workspace:^"
+  }
+}
--- a/plugins/refetch-interval/src/index.ts
+++ b/plugins/refetch-interval/src/index.ts
@@ -1,0 +1,141 @@
+import type { PiniaColadaPluginContext } from '@pinia/colada'
+import { toValue } from 'vue'
+/**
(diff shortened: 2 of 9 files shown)


posva:
Great! Does this overlap with https://github.com/posva/pinia-colada/pull/97 ? I yes, I might need to merge the other one first and then refactor the two PRs. We can change the API of plugins with breaking changes to improve them

plugins/refetch-interval/package.json
```diff
@@ -0,0 +1,73 @@
+{
+  "name": "@pinia/colada-plugin-refetch-interval",
+  "type": "module",
+  "publishConfig": {
+    "access": "public"
+  },
+  "version": "0.0.1",
+  "description": "Refetch queries on a regular interval with Pinia Colada",
+  "author": {
```
posva, line 9: Feel free to add yourself if you want: https://docs.npmjs.com/cli/v11/configuring-npm/package-json#people-fields-author-contributors

plugins/refetch-interval/src/index.ts
```diff
@@ -0,0 +1,141 @@
...
+ * @module @pinia/colada-plugin-refetch-interval
+ */
+
+/**
+ * Options for the Pinia Colada Refetch Interval plugin.
+ */
+export interface RefetchIntervalOptions {
+  /**
+   * If set to a number, queries will continuously refetch at this frequency in milliseconds.
+   * If set to false, no automatic refetching will occur.
```
posva, line 13: What about setting it to `true` or `'auto'` and have them auto refresh when stale? https://github.com/posva/pinia-colada/pull/97

plugins/refetch-interval/src/index.ts
```diff
@@ -0,0 +1,141 @@
...
+      after(() => {
+        // Only set up new interval if the fetch was successful and the query is enabled
+        if (queryEntry.state.value.status === 'success' && toValue(queryEntry.options?.enabled)) {
+          const setupNextInterval = () => {
+            const entry: IntervalEntry = {
+              lastUpdated: Date.now(),
+              timeoutId: setTimeout(() => {
+                // Only refetch if refetchIntervalInBackground is true or document is visible
+                if (options.refetchIntervalInBackground || isDocumentVisible) {
+                  Promise.resolve(queryCache.fetch(queryEntry))
```
posva, line 103: Does it make sense to force a fetch? If the data is fresh, it should be fine to not fetch again 🤔

plugins/refetch-interval/src/index.ts
```diff
@@ -0,0 +1,141 @@
...
+ */
+export function PiniaColadaRefetchInterval(
+  globalOptions?: RefetchIntervalOptions,
+): (context: PiniaColadaPluginContext) => void {
+  const defaults = { ...REFETCH_INTERVAL_OPTIONS_DEFAULTS, ...globalOptions }
+
+  return ({ queryCache }) => {
+    const intervalMap = new Map<string, IntervalEntry>()
+
+    let isDocumentVisible = typeof document !== 'undefined'
```
posva, line 45: I'm wondering if this makes sense at this level. It could be set in the `enabled` option so it feels a bit redundant here, the overlap between options could prove difficult in the future to debug.

plugins/refetch-interval/src/index.ts
```diff
@@ -0,0 +1,141 @@
...
+                    .finally(() => {
+                      // Setup next interval after completion
+                      setupNextInterval()
+                    })
+                } else {
+                  // If we can't refetch now, try again after the interval
+                  // setupNextInterval()
+
+                  // If we can't refetch now, invalidate the query
+                  queryCache.invalidate(queryEntry)
```
posva, line 114: FYI: invalidate also fetches if the query is active (displayed somewhere)

plugins/refetch-interval/src/refetchInterval.spec.ts
```diff
@@ -0,0 +1,518 @@
...
+ */
+import { enableAutoUnmount, flushPromises, mount } from '@vue/test-utils'
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+import { defineComponent } from 'vue'
+import { createPinia } from 'pinia'
+import { useQuery, PiniaColada } from '@pinia/colada'
+import type { PiniaColadaOptions, UseQueryOptions } from '@pinia/colada'
+import { PiniaColadaRefetchInterval } from '.'
+
+describe('Refetch Interval plugin', () => {
```
posva, line 12: Nice! A lot of tests 😄

posva:
I release the plugin `@pinia/colada-plugin-auto-refetch` if you want to try it out! I personally feel like the smart refresh based on `staleTime` is better. It still allows to configure the interval

vad1ym:
@posva Thanks for the review! As for the auto-refetch plugin, I didn’t realize right away that it basically does the same thing. I’ll give it a try, and it’ll probably be enough 👍

vad1ym:
@posva Auto-refetch plugin does everything I need, but I'm currently really missing the ability to control autoRefetch options via a ref or getter—for example, to disable auto-refetch when the user leaves the tab.

posva:
Nice catch! I will release a patch in a second.

posva:
I will close this for the moment, let me know if you find anything missing 🙂

Issue #44, opened 2024-05-16
add `useMutationsState()`


posva:
```ts
useMutationsState(filters: MutationFilters)
// or
useMutationState(opts: { filters: MutationFilters, select: (mutationEntry) => T })
```

It seems that it's often useful to select a part of the mutation entry. This also means that we first need the mutation store

shrpne:
please note, that the docs are mentioning `useMutationState` without `s`, not `useMutationsState`

Here:
- https://pinia-colada.esm.dev/api/@pinia/colada/interfaces/UseMutationOptions.html#key-
- https://pinia-colada.esm.dev/guide/optimistic-updates.html#When-mutation-is-not-collocated-with-the-query

posva:
I adapted the references to document how to implement your own

PR #189, opened 2025-02-18
refactor: make the return clearer

CaptainWang98:
Hi there,
I find it could be a clearer return structure while trying to do #44 ;
If I misunderstand your code, please let me know. It's my first step to contribute opensorce. 😊


posva:
Thanks but since there is a return statement I prefer just `if` since it's shorter

PR #204, opened 2025-02-28
fix: queryCache should not invalidate query when it disabled

### LINKED ISSUE #203: skip option for queries
There is enabled option, but actually "pauses" as written in docs

maybe I'm doing something wrong, but I really need skip option for example via plugin

> use case

I have defined query like this

```ts
export const useGetProject = defineQuery(() => {
  const route = useRoute();
  const projectId = computed(() => route.params.projectId as string);

  const { state, ...rest } = useQuery({
    key: () => [queryKeys.projects, projectId.value],
    query: () => apiGetProject(projectId.value),
    enabled: () => 'projectId' in route.params,
    skip: () => !projectId.value, // request example
  });

  return { ...rest, project: state };
});
```

and also mutation to update project

```ts
export const useUpdateProject = defineMutation({
  mutation: (data: ProjectUpdate) => apiUpdateProject(data),
  onSettled: () => {
    const queryCache = useQueryCache();
    queryCache.invalidateQueries({ key: [queryKeys.projects] });
  },
});
```

and I always getting something like `.../api/projects/undefined/` cause invalidation ignores enabled option

tednaaa:
closes #203

### DIFF
--- a/src/query-store.ts
+++ b/src/query-store.ts
@@ -444,7 +444,7 @@ export const useQueryCache = /* @__PURE__ */ defineStore(QUERY_STORE_ID, ({ acti
     return Promise.all(
       getEntries(filters).map((entry) => {
         invalidate(entry)
-        return entry.active && fetch(entry)
+        return entry.active && entry.options?.enabled && fetch(entry)
       }),
     )
   })
--- a/src/query-store.ts
+++ b/src/query-store.ts
@@ -444,7 +444,7 @@ export const useQueryCache = /* @__PURE__ */ defineStore(QUERY_STORE_ID, ({ acti
     return Promise.all(
       getEntries(filters).map((entry) => {
         invalidate(entry)
-        return entry.active && entry.options?.enabled && fetch(entry)
+        return entry.active && toValue(entry.options?.enabled) && fetch(entry)
       }),
     )
   })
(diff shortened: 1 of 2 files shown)


src/query-store.ts
```diff
@@ -444,7 +444,7 @@ export const useQueryCache = /* @__PURE__ */ defineStore(QUERY_STORE_ID, ({ acti
     return Promise.all(
       getEntries(filters).map((entry) => {
         invalidate(entry)
-        return entry.active && fetch(entry)
+        return entry.active && entry.options?.enabled && fetch(entry)
```
posva: Enabled can also be a ref, so we need this

```suggestion
        return entry.active && toValue(entry.options?.enabled) && fetch(entry)
```

src/use-query.spec.ts
```diff
@@ -997,6 +997,19 @@ describe('useQuery', () => {
     expect(query).toHaveBeenCalledTimes(2)
   })
 
+  it('should not invalidate query when enabled is false', async () => {
+    const { query } = mountSimple({ enabled: false })
```
posva, line 1001: let's add another test that has `() => false` and also skips fetching

posva (changes requested):

posva (approved):
Thanks!

PR #215, opened 2025-03-13
docs: clarify code example

onmax:
See https://github.com/posva/pinia-colada/discussions/214#discussioncomment-12488874


posva:
Thanks but this is actually fine!

PR #224, opened 2025-03-18
fix: make options parameter optional in types

43081j:
The types inside vue define that `Plugin<T>` where `T` isn't an array implies that the options _must_ be passed.

When you do specify a tuple/array, you're defining the parameters of the function. So `Plugin<[T0, T1]>` means `use(fn, T0, T1)`.

We can use this with an optional tuple member to make the parameter optional: `Plugin<[T?]>` (which is basically `(...params: [T?])`).

### DIFF
--- a/src/pinia-colada.ts
+++ b/src/pinia-colada.ts
@@ -29,7 +29,7 @@ export interface PiniaColadaOptions extends UseQueryOptionsGlobal {
  * @param app - Vue App
  * @param options - Pinia Colada options
  */
-export const PiniaColada: Plugin<PiniaColadaOptions> = (
+export const PiniaColada: Plugin<[PiniaColadaOptions?]> = (
   app: App,
   options: PiniaColadaOptions = {},
 ): void => {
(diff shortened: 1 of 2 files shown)


src/pinia-colada.ts
```diff
@@ -29,7 +29,7 @@ export interface PiniaColadaOptions extends UseQueryOptionsGlobal {
  * @param app - Vue App
  * @param options - Pinia Colada options
  */
-export const PiniaColada: Plugin<PiniaColadaOptions> = (
+export const PiniaColada: Plugin<[PiniaColadaOptions?]> = (
```
posva:
```suggestion
export const PiniaColada: Plugin<[options?: PiniaColadaOptions]> = (
```

posva (approved):
Thanks! I will wait for https://github.com/vuejs/core/pull/13063 to be merged and released as without it there is no auto completion in the options

43081j:
FYI there was a new release of vue so i think the referenced PR has now been released

do we still need this one?

PR #248, opened 2025-04-02
docs: mention retry plugin in tvq migration guide

87xie:
I used the retry plugin when migrating from TanStack Query to Pinia Colada. Mentioning this difference in the migration guide might be helpful.

Additionally, I would like to write some test cases for the retry plugin. Would it be okay to submit a PR for that? Thank you!

## Summary by CodeRabbit

- **Documentation**
  - Enhanced the migration guide with updated instructions on implementing retry functionality. Users now receive guidance on leveraging a dedicated plugin to enable reliable retry logic during the migration process.


posva:
Of course! Glad to have more tests!

posva (approved):

PR #249, opened 2025-04-03
fix(plugins): avoid retries when query is inactive

87xie:
This PR was originally intended to add test cases for the retry plugin. However, I noticed that the retry plugin still attempts retries even when the query is inactive, so I added a check for the inactive state. Feel free to close this if it’s not suitable.

Thanks for reviewing!

## Summary by CodeRabbit

- **Tests**
  - Introduced comprehensive tests covering various retry scenarios, including default settings, custom retry counts, conditional retries, and delay configurations.
  - Verified retry behavior stops on success, zero retries, or component unmounting, and resets on manual refetch.
  - Enhanced test reliability with automated setup, teardown, and use of fake timers.

- **Bug Fixes**
  - Prevented retry attempts on inactive queries to avoid unnecessary operations.

### DIFF
--- a/plugins/retry/src/index.ts
+++ b/plugins/retry/src/index.ts
@@ -113,6 +113,7 @@ export function PiniaColadaRetry(
           if (shouldRetry) {
             const delayTime = typeof delay === 'function' ? delay(entry.retryCount) : delay
             entry.timeoutId = setTimeout(() => {
+              if (!queryEntry.active) return
               // NOTE: we could add some default error handler
               isInternalCall = true
               Promise.resolve(queryCache.fetch(queryEntry)).catch(
--- a/plugins/retry/src/index.ts
+++ b/plugins/retry/src/index.ts
@@ -113,7 +113,10 @@ export function PiniaColadaRetry(
           if (shouldRetry) {
             const delayTime = typeof delay === 'function' ? delay(entry.retryCount) : delay
             entry.timeoutId = setTimeout(() => {
-              if (!queryEntry.active) return
+              if (!queryEntry.active) {
+                retryMap.delete(key)
+                return
+              }
               // NOTE: we could add some default error handler
               isInternalCall = true
               Promise.resolve(queryCache.fetch(queryEntry)).catch(
--- a/plugins/retry/src/retry.spec.ts
+++ b/plugins/retry/src/retry.spec.ts
@@ -1,7 +1,154 @@
-import { describe, expect, it } from 'vitest'
+/**
+ * @vitest-environment happy-dom
+ */
+import { enableAutoUnmount, flushPromises, mount } from '@vue/test-utils'
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+import { defineComponent } from 'vue'
+import { createPinia } from 'pinia'
+import { useQuery, PiniaColada } from '@pinia/colada'
+import type { UseQueryOptions } from '@pinia/colada'
+import { PiniaColadaRetry } from '.'
+import type { RetryOptions } from '.'
 
 describe('Pinia Colada Retry Plugin', () => {
-  it('works', () => {
-    expect(true).toBe(true)
+  beforeEach(() => {
+    vi.clearAllTimers()
+    vi.useFakeTimers()
+  })
+  afterEach(() => {
+    vi.restoreAllMocks()
+  })
+
+  enableAutoUnmount(afterEach)
+
+  function factory(
+    queryOptions: UseQueryOptions,
+    retryOptionsDefault?: RetryOptions,
+  ) {
+    const wrapper = mount(
+      defineComponent({
+        template: '<div></div>',
+        setup() {
+          return useQuery(queryOptions)
+        },
+      }),
+      {
+        global: {
+          plugins: [
+            createPinia(),
+            [
+              PiniaColada,
+              { plugins: [PiniaColadaRetry(retryOptionsDefault)] },
+            ],
+          ],
+        },
+      },
+    )
+
+    return { wrapper }
+  }
+
+  it('apply the default retry options', async () => {
+    const query = vi.fn(async () => {
+      throw new Error('ko')
+    })
+    const retryOptionsDefault = { retry: 1, delay: 200 }
+
+    factory({
+      key: ['key'],
+      query,
+    }, retryOptionsDefault)
+
+    // initial fetch fails
+    await flushPromises()
+    expect(query).toHaveBeenCalledTimes(1)
+    // first retry
+    vi.advanceTimersByTime(retryOptionsDefault.delay)
+    await flushPromises()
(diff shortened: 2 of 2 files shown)


plugins/retry/src/index.ts
```diff
@@ -113,6 +113,7 @@ export function PiniaColadaRetry(
           if (shouldRetry) {
             const delayTime = typeof delay === 'function' ? delay(entry.retryCount) : delay
             entry.timeoutId = setTimeout(() => {
+              if (!queryEntry.active) return
```
posva: shouldn't it also delete from the retryMap in this case to reset the count?
It would be great to add a failing test case first where the query with the same key is mounted again and see that the retry behavior is reset 🤔

posva (changes requested):

plugins/retry/src/index.ts
87xie: Thanks for pointing that out! It makes sense to delete it from the retryMap to reset the count.

plugins/retry/src/retry.spec.ts
```diff
@@ -1,7 +1,283 @@
-import { describe, expect, it } from 'vitest'
+/**
+ * @vitest-environment happy-dom
```
posva, line 2: Is this needed? Tests seems to pass without it

posva: I think I had it in delay but it's also not needed anymore. Nevermind the comment, I will remove them after merging 🙂

plugins/retry/src/retry.spec.ts
```diff
@@ -1,7 +1,283 @@
...
+    await flushPromises()
+    expect(query).toHaveBeenCalledTimes(3)
+  })
+
+  it('retry option with a function', async () => {
+    const query = vi.fn(async () => {
+      throw new Error('ko')
+    })
+    // stop retry when the error message is "ko"
+    const retry: RetryOptions['retry'] = (attempt, error) => {
```
posva: @coderabbitai can you inline this `retry` function into the option with the same name passed to `factory()`

plugins/retry/src/retry.spec.ts
```diff
@@ -1,7 +1,283 @@
...
+    vi.advanceTimersByTime(RETRY_OPTIONS_DEFAULTS.delay)
+    await flushPromises()
+    expect(query).toHaveBeenCalledTimes(1)
+  })
+
+  it('delay option with a number', async () => {
+    const query = vi.fn(async () => {
+      throw new Error('ko')
+    })
+    const delay = 500
```
posva: @coderabbitai inline this option

plugins/retry/src/retry.spec.ts
```diff
@@ -1,7 +1,283 @@
...
+    await flushPromises()
+    expect(query).toHaveBeenCalledTimes(2)
+  })
+
+  it('delay option with a function', async () => {
+    const query = vi.fn(async () => {
+      throw new Error('ko')
+    })
+    // 500ms for first retry, 1500ms for subsequent retries
+    const delay: RetryOptions['delay'] = (attempt) => {
```
posva: @coderabbitai inline this option, remove the comment above and format the option named `retry`

plugins/retry/src/retry.spec.ts
```diff
@@ -1,7 +1,283 @@
...
+    })
+    //  initial fetch fails
+    await flushPromises()
+    expect(query).toHaveBeenCalledTimes(1)
+    // first retry after 500ms
+    vi.advanceTimersByTime(500)
+    await flushPromises()
+    expect(query).toHaveBeenCalledTimes(2)
+    // second retry after 1500ms
+    vi.advanceTimersByTime(1500)
```
posva, line 166: @coderabbitai let's split this timer advance in two: let's advance 1000 first, check that the query was still called 2 times, then advance 500 to reach 1500 and these that the query is called 3 times

plugins/retry/src/retry.spec.ts
posva: can you create a PR to this PR with all the changes you replied in this review?

plugins/retry/src/retry.spec.ts
posva: use a `suggestion` tag instead of `diff` so that I can apply the change

posva: fix the indentation

plugins/retry/src/retry.spec.ts
```diff
@@ -1,7 +1,283 @@
...
+    const query = vi.fn(async () => {
+      throw new Error('ko')
+    })
+    const delay = 500
+
+    factory({
+      key: ['key'],
+      query,
+      retry: { delay },
+    })
```
posva: inline this `retry` function into the option with the same name passed to `factory()`. Use a suggestion instead of a diff

plugins/retry/src/retry.spec.ts
```diff
@@ -1,7 +1,283 @@
...
+    // 500ms for first retry, 1500ms for subsequent retries
+    const delay: RetryOptions['delay'] = (attempt) => {
+      return attempt === 0 ? 500 : 1500
+    }
+
+    factory({
+      key: ['key'],
+      query,
+      retry: { retry: 2, delay },
+    })
```
posva: @coderabbitai can you inline this option passed to `factory()`. Remove redundant comments. Use suggestions instead of diff

plugins/retry/src/retry.spec.ts
posva, line 166: let's split this timer advance in two: let's advance 1000 first, check that the query was still called 2 times, then advance 500 to reach 1500 and these that the query is called 3 times. Use suggestions instead of diff

posva (changes requested):
@coderabbitai fix review changes

posva:
@coderabbitai address the review comments

plugins/retry/src/retry.spec.ts
posva: @coderabbitai apply this and all the other comments I left in my review

plugins/retry/src/retry.spec.ts
posva: @coderabbitai address the comment above

posva: You are adding too much. The suggestion must contain only the changed lines

plugins/retry/src/retry.spec.ts
posva:
```suggestion
    // advance 1000ms (not enough time for next retry)
    vi.advanceTimersByTime(1000)
    await flushPromises()
    expect(query).toHaveBeenCalledTimes(2)
    // advance remaining 500ms to trigger second retry
    vi.advanceTimersByTime(500)
    await flushPromises()
    expect(query).toHaveBeenCalledTimes(3)
```

posva (approved):

PR #266, opened 2025-04-26
feat: add ResizablePanel component

nemanjamalesija:
This PR adds a reusable ResizablePanel component that currently supports vertical resizing.

There’s also an issue with wrapping the PipContainer inside the component. Wrapping the main content works fine, but the PipContainer isn’t behaving as expected and is left hanging in the background.

What’s done so far:

Basic ResizablePanel component is set up.
Vertical resizing works when wrapping any component except PipContainer.

What’s next:
Figure out how to properly wrap the PipContainer.
If needed, add support for horizontal resizing.

### DIFF
--- a/devtools/src/panel/DevtoolsPanel.ce.vue
+++ b/devtools/src/panel/DevtoolsPanel.ce.vue
@@ -3,6 +3,8 @@ import { ref, watch, onMounted, onUnmounted, provide } from 'vue'
 import type { UseQueryEntryPayload, DevtoolsEmits, AppEmits } from '@pinia/colada-devtools/shared'
 import { DuplexChannel } from '@pinia/colada-devtools/shared'
 import { DUPLEX_CHANNEL_KEY, QUERIES_KEY } from './composables/duplex-channel'
+import ResizablePanel from './components/ResizablePanel.vue'
+
 
 const { port, isPip } = defineProps<{
   port: MessagePort
@@ -63,6 +65,8 @@ channel.on('queries:delete', (q) => {
 
 <template>
   <PiPContainer id="root" :is-pip>
+    <!-- How to wrap ResizePanel around PiPContainer? -->
+   <ResizablePanel>
     <main class="w-full h-full grid grid-rows-[auto_1fr] bg-ui-bg text-ui-text font-sans">
       <!-- Merged Header with Tabs Navigation -->
       <div class="flex items-center border-b border-(--ui-border) select-none">
@@ -106,6 +110,7 @@ channel.on('queries:delete', (q) => {
 
       <RouterView />
     </main>
+    </ResizablePanel>
   </PiPContainer>
 </template>
 
--- a/devtools/src/panel/components/ResizablePanel.vue
+++ b/devtools/src/panel/components/ResizablePanel.vue
@@ -1,0 +1,83 @@
+
+<script setup lang="ts">
+import { ref, onMounted } from 'vue'
+import { useResizeObserver } from '@vueuse/core'
+
+const convertRemToPixels = (rem: number) => {
+  return rem * Number.parseFloat(getComputedStyle(document.documentElement).fontSize)
+}
+
+const isResizing = ref(false)
+const panelRef = ref<HTMLElement | null>(null)
+const panelRefWidth = ref(0)
+const panelHeight = ref('450px')
+const position = 'bottom'
+
+const handleDragStart = (event: MouseEvent) => {
+  const currentTarget = event.currentTarget as HTMLElement
+  const panelElement = currentTarget.parentElement
+  if (!panelElement) return
+
+  isResizing.value = true
+  const { height } = panelElement.getBoundingClientRect()
+  const startY = event.clientY
+  let newSize = 0
+  const minHeight = convertRemToPixels(3.5)
+
+  const runDrag = (moveEvent: MouseEvent) => {
+    moveEvent.preventDefault()
+
+      const valToAdd = startY - moveEvent.clientY
+      newSize = Math.round(height + valToAdd)
+      if (newSize < minHeight) {
+        newSize = minHeight
+      }
+
+      panelHeight.value = `${newSize}px`
+  }
+
+  const unsubscribe = () => {
+    if (isResizing.value) {
+      isResizing.value = false
+    }
+    document.removeEventListener('mousemove', runDrag, false)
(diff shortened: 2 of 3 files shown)


devtools/src/panel/DevtoolsPanel.ce.vue
```diff
@@ -3,6 +3,7 @@ import { ref, watch, onMounted, onUnmounted, provide } from 'vue'
 import type { UseQueryEntryPayload, DevtoolsEmits, AppEmits } from '@pinia/colada-devtools/shared'
 import { DuplexChannel } from '@pinia/colada-devtools/shared'
 import { DUPLEX_CHANNEL_KEY, QUERIES_KEY } from './composables/duplex-channel'
+import ResizablePanel from './components/ResizablePanel.vue'
```
posva: If it's in components/, it should get auto imported

devtools/src/panel/components/ResizablePanel.vue
```diff
@@ -0,0 +1,82 @@
+<script setup lang="ts">
+import { ref, onMounted } from 'vue'
+import { useResizeObserver } from '@vueuse/core'
+
+const convertRemToPixels = (rem: number) => {
```
posva: Always prefer named functions when possible. ~~I will add an eslint rule~~ I couldn't find one.

devtools/src/panel/components/ResizablePanel.vue
```diff
@@ -0,0 +1,82 @@
...
+  return rem * Number.parseFloat(getComputedStyle(document.documentElement).fontSize)
+}
+
+const isResizing = ref(false)
+const panelRef = ref<HTMLElement | null>(null)
```
posva: `useTemplateRef()` is the newer version of this

devtools/src/panel/components/ResizablePanel.vue
```diff
@@ -0,0 +1,82 @@
...
+    const { width } = entry.contentRect
+    panelRefWidth.value = width
+  })
+})
+</script>
+
+<template>
+  <div
+    ref="panelRef"
+    class="blabla fixed left-0 right-0 p-4 shadow-lg rounded-lg bg-gray-900   z-50 overflow-auto"
```
posva: Make sure to remove the extra spaces and unused classes 😄

devtools/src/panel/components/ResizablePanel.vue
```diff
@@ -0,0 +1,82 @@
...
+    :style="{
+      height: panelHeight,
+      width: '100vw',
+      top: 'auto',
+      bottom: position === 'bottom' ? '0' : 'auto',
+    }"
+    aria-label="Pinia colada devtools"
```
posva: probably not the right place to be

devtools/src/panel/components/ResizablePanel.vue
```diff
@@ -0,0 +1,82 @@
...
+})
+</script>
+
+<template>
+  <div
+    ref="panelRef"
+    class="blabla fixed left-0 right-0 p-4 shadow-lg rounded-lg bg-gray-900   z-50 overflow-auto"
+    :style="{
+      height: panelHeight,
+      width: '100vw',
```
posva: fixed values can go in classes

devtools/src/panel/components/ResizablePanel.vue
```diff
@@ -0,0 +1,82 @@
...
+      top: 'auto',
```
posva: auto is the initial value of top so this shouldn't do anything

devtools/src/panel/components/ResizablePanel.vue
```diff
@@ -0,0 +1,82 @@
...
+    }
+    document.removeEventListener('mousemove', runDrag, false)
+    document.removeEventListener('mouseup', unsubscribe, false)
+  }
+
+  document.addEventListener('mousemove', runDrag, false)
+  document.addEventListener('mouseup', unsubscribe, false)
+}
+
+onMounted(() => {
```
posva: No need for onMounted because these components are client only

devtools/src/panel/DevtoolsPanel.ce.vue
```diff
@@ -63,49 +64,52 @@ channel.on('queries:delete', (q) => {
...
 <template>
   <PiPContainer id="root" :is-pip>
-    <main class="w-full h-full grid grid-rows-[auto_1fr] bg-ui-bg text-ui-text font-sans">
-      <!-- Merged Header with Tabs Navigation -->
-      <div class="flex items-center border-b border-(--ui-border) select-none">
-        <!-- Logo -->
-        <div class="flex items-center p-2 mr-2">
-          <span class="text-xl">🍹</span>
-        </div>
+    <!-- How to wrap ResizePanel around PiPContainer? -->
```
posva: Maybe this ResizablePanel should just be part of PiPContainer

posva (commented):
I left some comments but overall we want to do something similar to https://github.com/vuejs/devtools/blob/main/packages/overlay/src/components/FrameBox.vue in vue devtools. The main difference is that here, it can only be resized from one edge rather than 4 different ones

devtools/src/panel/components/PiPContainer.vue
```diff
@@ -1,8 +1,10 @@
 <script setup lang="ts">
-import { useColorMode } from '@vueuse/core'
+import { useColorMode, useLocalStorage } from '@vueuse/core'
 import { computed } from 'vue'
+import ResizablePanel from './ResizablePanel.vue'
 
 const { isPip } = defineProps<{
+  id: string
```
posva: The id is needed for CSS.
Instead, adding `v-bind="$attrs"` to the aside is the way. This keeps this inheritance of attributes with a multi root node component

devtools/src/panel/components/ResizablePanel.vue
```diff
@@ -0,0 +1,62 @@
...
+import { useEventListener } from '@vueuse/core'
+
+const props = defineProps({
+  translate: {
+    type: Number,
+    required: true,
+  },
+})
+
+const emit = defineEmits(['update:translate'])
```
posva: use defineModel instead of props + emits

posva (approved):
I'm refactoring the devtools to be within the vue devtools so I fixed a couple of things to merge it

PR #273, opened 2025-04-30
feat(ui): control orientation of query splitpanes on small screens

nemanjamalesija:
Change orientation of query splitpanes below 800px

## Summary by CodeRabbit

- **New Features**
	- The layout of split panes now automatically adjusts orientation based on screen width, providing a horizontal split on smaller screens and a vertical split on larger screens for improved responsiveness.


devtools/src/panel/pages/queries.vue
```diff
@@ -81,7 +84,7 @@ const queriesGrouped = computed<
       </div>
     </div>
 
-    <Splitpanes class="overflow-hidden">
+    <Splitpanes class="overflow-hidden" :horizontal="screenWidth < 800">
```
nemanjamalesija, line 87: @posva should we move the breakpoint value (800 here) to some constant so it is not hardcoded like this?

nemanjamalesija:
will be implemented through separate pull request

Issue #267, opened 2025-04-29
Query makes multiple network requests for the same key in v0.15.1


vladyslav-mikhieiev:
In @pinia/colada version 0.15.1, when multiple components rendered using v-for call a composable created with defineQuery (which uses a static query key), a network request is triggered for each component instance upon mounting. This occurs even if the data for that key should be available in the cache and the staleTime has not expired.
This contrasts with the behavior in v0.14.2, where only the first component instance calling the composable would trigger the network request, and subsequent components would correctly receive the data from the cache (deduplication worked as expected).
This behavior leads to unnecessary network requests and potentially impacts performance, especially with large lists. We have observed this pattern not only in the specific scenario described below but also in other parts of our application where different components access the same query data (via defineQuery-based composables) that should be cached.

[0.14.2 Demo](https://stackblitz.com/edit/github-mnspnpcg-edsezyhx?file=src%2Fcomponents%2FArtworkCard.vue)

[0.15.1 Demo](https://stackblitz.com/edit/github-mnspnpcg?file=src%2Fcomponents%2FArtworkCard.vue)

![Image](https://github.com/user-attachments/assets/dd243228-0a01-4639-b472-ac58df0b03fb)

posva:
Thanks for the report. If you could you boil it down to a unit test and open  PR, that would be perfect!

vladyslav-mikhieiev:
> Thanks for the report. If you could you boil it down to a unit test and open PR, that would be perfect!

Ok, I'll try.

vladyslav-mikhieiev:
I tried writing a test to track repeated requests, but I was only ever able to observe a single call to the request function.

Then I decided to look into the commit history between versions to find the root cause — and I think I found it.

There was a commit: [feat(hmr): refetch on component change.](https://github.com/posva/pinia-colada/commit/56aad7aa268d90d17824837342e6260968216d8c)
This logic is what triggers the repeated requests in the scenario described in the issue.

Notably, when running the app with NODE_ENV=production, the repeated requests do not occur.

I’m not entirely sure what to do about it.
@posva, any thoughts?

posva:
It could be an HMR bug. I think it should still be reproductible without data loaders

posva:
[...]

PR #279, opened 2025-05-10
fix(query): enhance HMR handling

vladyslav-mikhieiev:
Fixed HMR issue with duplicate network requests in multiple component instances

- Core: Enhanced HMR tracking to use composite keys that include component instance UID
- Core: Updated component instance tracking in both ensure and untrack functions
- Core: Improved UseQueryEntry type documentation for __hmr property

This PR fixes issue [#267](https://github.com/posva/pinia-colada/issues/267) where multiple instances of the same component (e.g., in v-for loops) were causing duplicate network requests during HMR. The problem occurred because HMR tracking was only using component type __hmrId but not distinguishing between different instances of the same component.

The solution implements composite keys using both the component type __hmrId and the instance uid to uniquely identify each component instance, ensuring proper invalidation tracking during hot module replacement.

This change only affects development mode and has no impact on production builds.

## Summary by CodeRabbit

- **Documentation**
  - Updated internal documentation to clarify how component instances are tracked during hot module replacement.
- **Bug Fixes**
  - Improved tracking of component instances for hot module replacement to better distinguish between different instances of the same component type.

### DIFF
--- a/src/query-store.ts
+++ b/src/query-store.ts
@@ -134,6 +134,7 @@ export interface UseQueryEntry<
   __hmr?: {
     /**
      * Reference count of the components using this query.
+     * Key is a composite key of component hmrId and instance uid.
      */
     ids: Map<string, number>
   }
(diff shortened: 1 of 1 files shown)


posva (changes requested):
Thanks! Could you add a test based on the original issue code (without data loaders)? 
I remember you said you couldn't write a failing test. Could you still share what you got? I might be able to help

vladyslav-mikhieiev:
I tested something slightly different from what was actually needed. Now I think I need to update the old HMR test and add more, but I’ve never worked with such tests before and don’t have time to figure it out right now. I’d appreciate it if you could handle the test part. @posva 

I did test it locally in my project, and there’s no bug after the changes.

posva:
No worries. I was trying to reproduce it but I also couldn't. It turns out it is fixed in main. I will check if this change is needed

PR #286, opened 2025-05-19
docs: use getter function for key option in SSR examples

87xie:
It looks like we should use a getter function for the `key` option.


posva (approved):
Thanks!

PR #297, opened 2025-05-26
feat: merge devtools into main package

posva:
I'm still unsure of this as it doesn't really work well with Vite. Having an external component that the user manually includes is still the best option.


stefanvanherwijnen:
Would it be possible to create multiple builds, e.g. a `production` build in which @vue/devtools-api is not included?
There is a check for it in the source: https://github.com/posva/pinia-colada/blob/4f5dc12ad0f138a8a524ef1ecb67b61aa8535f12/src/pinia-colada.ts#L71

However, it is always statically imported in the distributed `index.js`: 
```ts
import { setupDevtoolsPlugin } from "@vue/devtools-api";
```
Which in my case leads to a broken bundle.

posva:
Moved to a custom version of devtools in the end for more power

PR #303, opened 2025-06-01
chore: update name of the entity being documented

ackzell:
I noticed this while hovering over the api method being imported.

image

## Summary by CodeRabbit

- **Documentation**
  - Updated comments to accurately refer to products instead of contacts in the product API.


posva (approved):
Thanks!

PR #343, opened 2025-07-30
fix: make type inference of DefineQueryOptions.placeholderData correct

### LINKED ISSUE #337: placeholderData inferred as "any" in defineQuery and defineQueryOptions
`placeholderData` type inference works differently from `useQuery` where it works as expected:

Image

CaptainWang98:
Fix #337 
This may caused by `&` in typescript. I try use `Omit` and it works.
But I'm not sure what `we need to duplicate the types for initialData and placeholderData` means, can @posva explain plz?
https://github.com/posva/pinia-colada/blob/acefffa398d144c51826fd0e41c953f5a4de55e3/src/define-query.ts#L25-L40

## Summary by CodeRabbit

* **Refactor**
  * Clarified handling of initial vs. placeholder query data: initial data now explicitly establishes a successful query state, while placeholder data can be provided via a callback that derives a transient value from prior data without mutating cache, yielding more predictable loading and status behavior.
* **Tests**
  * Added type-level tests verifying placeholder-data callbacks receive previous data (or undefined) and return expected fallback values.


posva (changes requested):
While I was playing around with this idea, I found that it breaks other stuff.
There are some commented out tests that need to be uncommented as well as new dts tests and docs copied (duplicated as mentioned in the comments below)

CaptainWang98:
> While I was playing around with this idea, I found that it breaks other stuff. There are some commented out tests that need to be uncommented as well as new dts tests and docs copied (duplicated as mentioned in the comments below)

Could u plz show some example about `it breaks other stuff`?
What's your expect result for #337 , plz let me know and I'll write some test case first, then try to figure out better approach to fix it. 😊

posva:
Sure! check the tests at https://github.com/posva/pinia-colada/blob/main/src/use-query.test-d.ts#L211. You will see placeholderData tests are missing at https://github.com/posva/pinia-colada/blob/main/src/define-query.test-d.ts and https://github.com/posva/pinia-colada/blob/main/src/define-query-options.test-d.ts. Add the tests to ensure everything work

src/define-query.ts
```diff
@@ -26,7 +26,7 @@ export type DefineQueryOptions<
   TData = unknown,
   TError = ErrorDefault,
   TDataInitial extends TData | undefined = undefined,
-> = _RemoveMaybeRef<UseQueryOptions<TData, TError, TDataInitial>> & {
+> = Omit<_RemoveMaybeRef<UseQueryOptions<TData, TError, TDataInitial>>, 'initialData' | 'placeholderData'> & {
   // NOTE: we need to duplicate the types for initialData and placeholderData to make everything work
   // we omit the descriptions because they are inherited from the original type
```
posva: If the type is omitted, we need to copy the docs from the original type

CaptainWang98: This caused by using both `_RemoveMaybeRef` and `&` here. The issue disappears if you remove one of them.
It may related to unexpected property merge strategy.
https://github.com/posva/pinia-colada/blob/e9b9972d4f47dbb69533926bb89de0b03c5535f0/src/define-query.ts#L29

posva (approved):
Thanks! I think it's safe to merge this now. I suspect I might need to remove the helper in the future to make things better 😆

PR #361, opened 2025-08-07
docs: clarify gc behavior of `setQueryData`

87xie:
Added a note about GC behavior in `setQueryData`. I'm not sure it's necessary, so feel free to close if it's not needed.

## Summary by CodeRabbit

* **Documentation**
  * Updated documentation to clarify how unused cached query data is automatically removed after a configurable period.

### DIFF
--- a/docs/cookbook/prefetching.md
+++ b/docs/cookbook/prefetching.md
@@ -36,3 +36,5 @@ queryCache.setQueryData(
   await fetch('/api/users').then((res) => res.json()),
 )
 ```
+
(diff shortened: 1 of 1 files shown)


posva (commented):
Why add it here. I don't think it matches the page. I don't know if there is a correct place to put this yet

87xie:
> Why add it here. I don't think it matches the page. I don't know if there is a correct place to put this yet

At first glance, `setQueryData` doesn’t expose any query options like `gcTime`, so I wasn’t sure whether the data would still be subject to garbage collection (similar to TanStack Query).

I agree this isn’t the right place, so feel free to close the PR

posva:
Still thanks a lot for opening the pr

PR #376, opened 2025-08-29
feat: add ssrThrowOnError

WellDone2094:
Add ssrThrowOnError query option to allow the user decide if during ssr the errors should be propagated or not


src/query-options.ts
```diff
@@ -62,6 +62,12 @@ export interface UseQueryOptionsGlobal {
    * with `initialData`, the placeholder does not change the cache state.
    */
   placeholderData?: (previousData: unknown) => any // any allows us to not worry about the types when merging options
+
+  /**
+   * Whether to throw errors during SSR (onServerPrefetch) when the query fails.
+   * @default true
+   */
+  ssrThrowOnError?: boolean
```
posva: Let's rename to `ssrCatchError` so the default value is falsy

src/ssr.spec.ts
```diff
@@ -80,5 +80,37 @@ describe('SSR', () => {
...
+      appSetup() {
+        onErrorCaptured(spy)
+      },
+    })
+    query.mockRejectedValueOnce(new Error('ko'))
+    expect(await renderToString(app)).toMatchInlineSnapshot(`"<!---->"`)
+    expect(spy).toHaveBeenCalledTimes(0)
+  })
+
   it.todo('can avoid throwing inside onServerPrefetch')
```
posva: This one can be removed now

src/ssr.spec.ts
```diff
@@ -67,7 +67,7 @@ describe('SSR', () => {
     }
   }
 
-  it('propagates query errors', async () => {
+  it('propagates query errors when using default query options', async () => {
```
posva:
```suggestion
  it('propagates query errors by default', async () => {
```

src/ssr.spec.ts
```diff
@@ -80,5 +80,37 @@ describe('SSR', () => {
     expect(spy).toHaveBeenCalledTimes(1)
   })
 
+  it('propagates query errors when ssrThrowOnError is true', async () => {
```
posva: this test is redundant and can be removed

src/ssr.spec.ts
```diff
@@ -80,5 +80,37 @@ describe('SSR', () => {
...
+      appSetup() {
+        onErrorCaptured(spy)
+      },
+    })
+    query.mockRejectedValueOnce(new Error('ko'))
+    expect(await renderToString(app)).toMatchInlineSnapshot(`"<!---->"`)
+    expect(spy).toHaveBeenCalledTimes(1)
+  })
+
+  it('Don\'t propagates query errors when ssrThrowOnError is false', async () => {
```
posva:
```suggestion
  it(`Don't propagates query errors when ssrThrowOnError is false`, async () => {
```

src/use-query.ts
```diff
@@ -286,7 +286,7 @@ export function useQuery<
   if (hasCurrentInstance) {
     // only happens on server, app awaits this
     onServerPrefetch(async () => {
-      if (toValue(enabled)) await refresh(true)
+      if (toValue(enabled)) await refresh(options.value.ssrThrowOnError ?? optionDefaults.ssrThrowOnError)
```
posva: options.value already include the default, the `??` fallback case shouldnt' be needed

src/query-options.ts
```diff
@@ -169,6 +176,7 @@ export const USE_QUERY_DEFAULTS = {
   refetchOnReconnect: true as NonNullable<UseQueryOptions['refetchOnReconnect']>,
   refetchOnMount: true as NonNullable<UseQueryOptions['refetchOnMount']>,
   enabled: true as MaybeRefOrGetter<boolean>,
+  ssrThrowOnError: true as NonNullable<UseQueryOptions['ssrThrowOnError']>,
```
posva: with the change of name we can avoid adding this default too

posva (changes requested):
Thanks!

src/query-options.ts
```diff
@@ -62,6 +62,18 @@ export interface UseQueryOptionsGlobal {
    * with `initialData`, the placeholder does not change the cache state.
    */
   placeholderData?: (previousData: unknown) => any // any allows us to not worry about the types when merging options
+
+  /**
+   * Whether to catch errors during SSR (onServerPrefetch) when the query fails.
+   * @default true
```
posva:
```suggestion
   * @default false
```

src/query-options.ts
```diff
@@ -62,6 +62,18 @@ export interface UseQueryOptionsGlobal {
...
+   */
+  ssrCatchError?: boolean
+
+  /**
+   * Whether to throw errors during SSR (onServerPrefetch) when the query fails.
+   * @default true
+   */
+  ssrThrowOnError?: boolean
```
posva:
```suggestion
```

posva (approved):

PR #377, opened 2025-09-03
feat: add ssrPrefetch query-option

WellDone2094:
## Summary by CodeRabbit

- New Features
  - Added an SSR option to control query prefetching during server-side rendering (ssrPrefetch). Enabled by default; can be disabled per query or via global defaults.

- Tests
  - Added coverage to verify that queries are not prefetched during SSR when ssrPrefetch is disabled.

### DIFF
--- a/src/query-options.ts
+++ b/src/query-options.ts
@@ -68,6 +68,12 @@ export interface UseQueryOptionsGlobal {
    * @default false
    */
   ssrCatchError?: boolean
+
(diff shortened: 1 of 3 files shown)


posva:
Thanks but you can disable the query on server with `enabled: !isServer`: https://pinia-colada.esm.dev/guide/ssr.html#Lazy-queries

PR #400, opened 2025-10-16
Export `toCacheKey`

jumika:
Fixed: toCacheKey is only reexported as a type

## Summary by CodeRabbit

* **Refactor**
  * Updated the public API export structure for improved consistency in how library identifiers are made available to consumers.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -12,7 +12,7 @@ export type {
   DataStateStatus,
 } from './data-state'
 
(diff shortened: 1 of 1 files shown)


jumika:
I think this change is quite self explanatory. The documentation mentions this function, but the implementation was not exported.
No breaking changes.

posva (approved):

PR #406, opened 2025-10-20
docs: refine query invalidation docs

87xie:
In our project, we have some cases where `predicate` is used, so I added a usage example to the docs and updated the wording for better consistency.

Hope this helps, and feel free to close this if it’s not suitable.

## Summary by CodeRabbit

* **Documentation**
  * Clarified query invalidation guide terminology and default behaviors
  * Added comprehensive examples, including predicate-based invalidation and end-to-end mutation flows
  * Introduced guidance on mutation workflows: onSettled usage, awaiting invalidation, and optimistic update patterns
  * Updated a documentation link target for the "Invalidate queries" entry under mutation guidance

### DIFF
--- a/docs/guide/query-invalidation.md
+++ b/docs/guide/query-invalidation.md
@@ -2,7 +2,7 @@
 
 Query invalidation is crucial for maintaining up-to-date data in your application. Typically, you should invalidate queries following mutations to ensure the cache stays synchronized with the server. However, you can also manually invalidate queries using the [Query Cache](../advanced/query-cache.md). Let's cover both methods.
 
-## Invalidation via Query Cache
+## Invalidation via the Query Cache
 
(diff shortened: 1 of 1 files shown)


docs/guide/query-invalidation.md
```diff
@@ -20,6 +20,16 @@ queryCache.invalidateQueries({ key: ['todos'], exact: true })
 queryCache.invalidateQueries()
 ```
 
+You can also use a `predicate` function for custom filtering logic. It receives a query entry and returns a boolean indicating whether the query should be invalidated:
+
+```ts
+// Invalidate queries that key starts with `users` or `todos`
+const resourcesToInvalidate = ['users', 'todos']
+queryCache.invalidateQueries({
+  predicate: (entry) => resourcesToInvalidate.includes(entry.key[0]),
```
posva: Let’s just use an if without the array + includes, I think it makes the example clearer

PR #407, opened 2025-10-22
fix(auto-refetch): prevent wrong entry refetch on reactive key change

### LINKED ISSUE #405: useQuery: autoRefetch executes once even when enabled changes from true to false
### Problem Description

When `useQuery`'s `enabled` option changes from `true` to `false`, the `autoRefetch` feature still executes one more time because the internal `setTimeout` is not properly cancelled when `enabled` becomes `false`.

### Expected Behavior

When `enabled` is set to `false`, autoRefetch should stop immediately and not execute any additional refetches.

fcsha:
Fix #405

When query options change reactively (e.g., key or enabled), the old entry's 
timer would refetch the NEW entry's data instead of its own. This happened 
because the timer callback re-queried entries using toValue(options.key), 
which always returns the current key value.

For example:
- useQuery({ key: userId, autoRefetch: 5000 })  // userId = 'user-1'
- Timer A scheduled for user-1 entry
- userId changes to 'user-2', creating a new entry
- Timer A fires and queries with current key 'user-2', fetching wrong entry's data

Changes:
- Use entry reference directly in setTimeout instead of re-querying by key
- Each timer now binds to its own entry and checks entry.active before refetch
- Add check for disabled queries to clear timeout in 'ensure' hook

## Summary by CodeRabbit

* **Bug Fixes**
  * Auto-refetch now correctly halts when a query's enabled flag is false and avoids redundant scheduling.
  * Timers are cleared appropriately on fetch/removal.

* **New Features**
  * Auto-refetch configuration is now available via query options.

* **Tests**
  * Added tests verifying enabled behavior and that changing a query key resets the refetch timer.


fcsha:
I've pushed the update. Is this alright?

posva:
@fcsha i think the pr auto closed

PR #416, opened 2025-10-28
fix(auto-refetch): prevent wrong entry refetch on reactive key change

### LINKED ISSUE #405: useQuery: autoRefetch executes once even when enabled changes from true to false

fcsha:
Fix #405

When query options change reactively (e.g., key or enabled), the old entry's
timer would refetch the NEW entry's data instead of its own. This happened
because the timer callback re-queried entries using toValue(options.key),
which always returns the current key value.

For example:
- useQuery({ key: userId, autoRefetch: 5000 }) // userId = 'user-1'
- Timer A scheduled for user-1 entry
- userId changes to 'user-2', creating a new entry
- Timer A fires and queries with current key 'user-2', fetching wrong entry's data

Changes:
- Use entry reference directly in setTimeout instead of re-querying by key
- Each timer now binds to its own entry and checks entry.active before refetch
- Add check for disabled queries to clear timeout in 'ensure' hook

## Summary by CodeRabbit

* **Bug Fixes**
  * Auto-refetch now respects the enabled option (pauses when disabled, clears scheduled refetch)
  * Refetch timer is reset when the query key changes, avoiding unexpected fetches

* **Tests**
  * Added tests validating enabled-state handling and pausing behavior
  * Added tests verifying timer reset and correct behavior after key changes

### DIFF
--- a/plugins/auto-refetch/src/auto-refetch.spec.ts
+++ b/plugins/auto-refetch/src/auto-refetch.spec.ts
@@ -193,40 +193,25 @@ describe('Auto Refetch plugin', () => {
   })
 
   it('respects query enabled option', async () => {
-    const enabled = ref(false)
+    const enabled = ref(true)
     const { query } = mountQuery({
       enabled,
       staleTime: 1000,
     })
 
-    // When query enabled = false, query should not be called initially
-    await flushPromises()
-    expect(query).toHaveBeenCalledTimes(0)
-
-    vi.advanceTimersByTime(2000)
-    await flushPromises()
-    // Still should not refetch because query is disabled
-    expect(query).toHaveBeenCalledTimes(0)
-
-    // Change query enabled to true
-    enabled.value = true
+    // Wait for initial query
     await flushPromises()
     // Now query should be called
     expect(query).toHaveBeenCalledTimes(1)
 
-    vi.advanceTimersByTime(1000)
-    await flushPromises()
-    // Should refetch because query is enabled and auto-refetch is active
-    expect(query).toHaveBeenCalledTimes(2)
-
-    // Change query enabled back to false
+    // Change query enabled to false
     enabled.value = false
     await flushPromises()
     // Advance time - should NOT refetch because query is disabled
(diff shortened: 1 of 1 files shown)


fcsha:
Replaces #407

The previous PR was closed due to an accidental fork sync on my end that wiped the branch. Re-submitting the same changes here. Sorry for the noise!

plugins/auto-refetch/src/auto-refetch.spec.ts
```diff
@@ -192,6 +192,75 @@ describe('Auto Refetch plugin', () => {
...
+    await flushPromises()
+    expect(query).toHaveBeenCalledTimes(0)
+
+    vi.advanceTimersByTime(2000)
+    await flushPromises()
+    // Still should not refetch because query is disabled
+    expect(query).toHaveBeenCalledTimes(0)
+
+    // Change query enabled to true
+    enabled.value = true
```
posva: Start directly with the enabled flag as true, the part before is already tested in use query and is not relevant to auto refetch

plugins/auto-refetch/src/auto-refetch.spec.ts
```diff
@@ -192,6 +192,75 @@ describe('Auto Refetch plugin', () => {
...
+    await flushPromises()
+    // Now query should be called
+    expect(query).toHaveBeenCalledTimes(1)
+
+    vi.advanceTimersByTime(1000)
+    await flushPromises()
+    // Should refetch because query is enabled and auto-refetch is active
+    expect(query).toHaveBeenCalledTimes(2)
```
posva: this is already tested by other tests, so isn't needed

PR #441, opened 2025-12-16
docs: simplify query function parameter descriptions

87xie:
According to the `useQuery` API, the `query` function signature is:

```ts
(context: UseQueryFnContext) => Promise<TData>
```

Some parts of the docs mention that the `query` function takes **no arguments**, which could be misleading, so I removed those descriptions since the code example below doesn’t use the `abortSignal` from the context parameter.

Feel free to close this if it is not suitable

## Summary by CodeRabbit

* **Documentation**
  * Clarified that query functions are not restricted to no-argument forms and can accept parameters and use external properties.
  * Updated guide and quick-start examples to reflect the expanded query capabilities.
  * Added note that query functions are automatically executed by the framework when needed.

### DIFF
--- a/docs/guide/queries.md
+++ b/docs/guide/queries.md
@@ -87,7 +87,7 @@ Most of the time you will find yourself using just `state` and `asyncStatus` to 
 
 ## Using External Properties in Queries
 
-Since queries are automatically triggered by Pinia Colada, you can directly use external properties like route params or search queries within the `query` function. To ensure proper caching, add these properties to the `key` as a function. A common example is using the `route` within the `query` function:
+Since queries are automatically triggered by Pinia Colada, the `query` function cannot accept custom parameters. However, you can directly use external properties like route params or search queries within the `query` function. To ensure proper caching, add these properties to the `key` as a function. A common example is using the `route` within the `query` function:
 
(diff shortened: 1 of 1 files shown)


docs/guide/queries.md
```diff
@@ -87,7 +87,7 @@ Most of the time you will find yourself using just `state` and `asyncStatus` to
 
 ## Using External Properties in Queries
 
-Since queries are automatically triggered by Pinia Colada, the `query` function cannot accept parameters. However, you can directly use external properties like route params or search queries within the `query` function. To ensure proper caching, add these properties to the `key` as a function. A common example is using the `route` within the `query` function:
+Since queries are automatically triggered by Pinia Colada, you can directly use external properties like route params or search queries within the `query` function. To ensure proper caching, add these properties to the `key` as a function. A common example is using the `route` within the `query` function:
```
posva:
```suggestion
Since queries are automatically triggered by Pinia Colada, the `query` function cannot accept custom parameters. However, you can directly use external properties like route params or search queries within the `query` function. To ensure proper caching, add these properties to the `key` as a function. A common example is using the `route` within the `query` function:
```

posva (approved):
Thanks!

Issue #442, opened 2025-12-17
Plugin added properties lost when assigned during `create` action


87xie:
When a plugin assigns extension properties during the `create` action, those properties may be lost later. 

This can be observed with the delay plugin. On a cache miss, `ensure` invokes `create` and initializes new entries with `ext: START_EXT`. Later in the same ensure flow, `START_EXT` is replaced with a fresh object:

```ts
let entry = cachesRaw.get(keyHash)

if (!entry) {
  cachesRaw.set(
    keyHash,
    (entry = create(...)),
  )
}

if (entry.ext === START_EXT) {
  entry.ext = {}
  extend(entry)
}
```

Since `create` is part of `ensure`, a plugin’s `after` handler for the `create` action may write to the shared `START_EXT` before it is replaced:

```ts
queryCache.$onAction(({ name, after }) => {
  if (name === 'create') {
    after((entry) => {
      scope.run(() => {
        const isDelaying = shallowRef(false)
        entry.ext.isDelaying = isDelaying
      })
    })
  }
})
```

As a result, properties previously attached to `entry.ext` are discarded, leading to runtime `undefined` reads:

```ts
const {
  isDelaying, // undefined
} = useQuery(...)
```

87xie:
I think this can be changed to assign properties in `extend` action, like:

```ts
if (name === 'extend') {
  after(() => {
    const [entry] = args
    scope.run(() => {
      const isDelaying = shallowRef(false)
      entry.ext.isDelaying = isDelaying
    })
  })
}

```

If that sounds reasonable, I can open a PR to update the plugin documentation and tests.

posva:
Sorry, I didn't see the message but it's exactly what you said, it **has** to be in `extend`

PR #443, opened 2025-12-17
docs: update the `dataUpdatedAt` plugin example

87xie:
Update the `dataUpdatedAt` plugin example to match the approach used in #442.

## Summary by CodeRabbit

* **Documentation**
  * Clarified plugin lifecycle behavior: specifies the correct hook used to add custom properties, explains when those properties become available, and notes scope constraints affecting timing and access within handlers.


posva (approved):
Oh, thanks! Nice catch

PR #464, opened 2026-01-14
fix: correct GC time display for prefetched queries

### LINKED ISSUE #460: Displayed incorrect GC time in DevTools for prefetched queries
I noticed that when a query is prefetched and has no observers, the GC timing information is displayed incorrectly in DevTools.

In our case, we prefetch certain queries after the login mutation, like:

```ts
const profileQueryOptions = defineQueryOptions({
  key: AUTH_KEYS.profile(),
  query: getProfile,
})

function useLoginMutation() {
  const cache = useQueryCache()

  return useMutation({
    mutation: login,
    onSuccess() {
      cache.refresh(cache.ensure(profileQueryOptions))
    },
  })
}
```

On some pages, the profile data is not consumed immediately (i.e. there are no active observers). In this scenario, DevTools shows the GC timing as “56 years ago"

Image

Root cause: 

1. `inactiveAt` is initialized to `0` for all query entries
2. Prefetched queries without observers never trigger `untrack` action
3. `new Date(selectedQuery.devtools.inactiveAt + selectedQuery.options.gcTime)` creates a timestamp from 1970

Not sure if using `updatedAt` as a fallback when `inactiveAt === 0` is the right way to calculate GC time

Thanks!

87xie:
Fixes #460

Both prefetch methods (`setQueryData` and `refresh`) call `setEntryState` when the request completes. Therefore, I updated `pc-devtools-info-plugin` to intercept the `setEntryState` action.

If a query entry is inactive after a state update, `inactiveAt` is now explicitly set to the current timestamp.

Note: The test file `issue-460.vue` was generated by AI and slightly refined, then tested. If this isn't suitable, feel free to remove it or close this PR.

Discussion: Would it make sense to centralize `scheduleGarbageCollection` inside `setEntryState`, since every state update should re-evaluate GC needs?

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected garbage-collection timing so prefetched queries reliably start inactive timing.

* **New Features**
  * Added a playground demo page to reproduce and observe prefetched query GC behavior.

* **Chores**
  * Updated routing/type info to include the new demo pages.

* **Refactor**
  * Centralized GC scheduling to run on entry state updates for more consistent behavior.

### DIFF
--- a/src/query-store.ts
+++ b/src/query-store.ts
@@ -730,8 +730,6 @@ export const useQueryCache = /* @__PURE__ */ defineStore(QUERY_STORE_ID, ({ acti
                 // reset the placeholder data to free up memory
                 entry.placeholderData = null
               }
-              // schedule GC for prefetched queries with no observers
-              scheduleGarbageCollection(entry)
             }
           }),
         when: Date.now(),
@@ -786,8 +784,9 @@ export const useQueryCache = /* @__PURE__ */ defineStore(QUERY_STORE_ID, ({ acti
     ) => {
       entry.state.value = state
       entry.when = Date.now()
-      // if we need to, we could schedule a garbage collection here but I don't
-      // see why would one create entries that are not used (not tracked immediately after)
+      // re-evaluate GC needs when the state is updated. One such scenario is
+      // prefetched queries that bypass track/untrack and would never be garbage collected otherwise
+      scheduleGarbageCollection(entry)
     },
   )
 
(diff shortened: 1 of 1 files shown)


posva:
> Discussion: Would it make sense to centralize scheduleGarbageCollection inside setEntryState, since every state update should re-evaluate GC needs?

I think yes, seting the entry state should reset any timer

87xie:
> Looks good to me! Feel free to add to this PR the refactor of scheduling gc. It might even allow to simplify the code a bit more

Thanks! I've included the GC scheduling refactor here to see if it helps simplify the code.

posva (approved):

Issue #487, opened 2026-02-07
Support `recentlySuccessful` (similar to Inertia)


Barbapapazes:
Hey 👋,

In order to display a small feedback to the user, without having to create a toast as they cause a lot of issues (https://www.reddit.com/r/UXDesign/comments/1p7fph6/github_retiring_toasts_from_their_design_system/), having a variable named `recentlySuccessful`, that is true when item is saved and then false after a couple of second could be really useful.

Similar to https://inertiajs.com/docs/v2/the-basics/forms#form-errors:~:text=forms%20have%20a-,recentlySuccessful,-property%2C%20which%20will

posva:
I remember your demo at Vue.js Paris. I personally don't like this practice because I think it can be confusing: you send a form, look away, someone asks you a question, look back, what happened? I thought I submitted the form. So I don't want to include it in the library itself or as a core plugin that I have to maintain later on.

I will be happy to add a link to a plugin that adds this feature (and other features if you want) to a new Plugins / community plugins section in docs

Barbapapazes:
I'll try to do a plugin and keep you informed about it!

Barbapapazes:
@posva do you think you could provide something similar to https://pinia-colada.esm.dev/advanced/plugins.html#Examples but for mutations? There is no examples nor existing plugins and I don't know how to add a new parameter. Thank you!

Barbapapazes:
I'm using something like

```ts
 return ({ pinia }) => {
    const mutationCache = useMutationCache(pinia)

    mutationCache.$onAction(({ name, args, after, onError }) => {
      if (name === 'extend') {
        const [entry] = args as [UseMutationEntry<any, any, any>]
        entry.ext.recentlySuccessful = shallowRef(false)
      }
}
```

but when using it, `recentlySuccessful` is undefined.

```ts
const { mutate, recentlySuccessful } = useMutation({
  mutation: () => new Promise((resolve) => {
    setTimeout(() => {
      console.log('Mutation completed');
      resolve('done');
    }, 1000);
  }),
})
```

Barbapapazes:
I suspect this line https://github.com/posva/pinia-colada/blob/71aa7f66d79a3658458b887b2f745ba411f7985b/src/use-query.ts#L272 to be missing in the `useMutation` return https://github.com/posva/pinia-colada/blob/71aa7f66d79a3658458b887b2f745ba411f7985b/src/use-mutation.ts#L205

posva:
[...]

PR #490, opened 2026-02-08
feat: plugin support for mutations

Barbapapazes:
related to https://github.com/posva/pinia-colada/issues/487#issuecomment-3867282968

## Summary by CodeRabbit

* **New Features**
  * Added a mutation metrics plugin that exposes mutatedAt (timestamp of last successful mutation) and errorCount (mutation error count) on mutations; useMutation return values now surface plugin-provided extensions.

* **Documentation**
  * Added guides and examples showing how to extend mutation options and entries and how to access the new mutation metrics in components.

* **Tests**
  * Added test coverage for the mutation metrics behavior (success and error paths) and for plugin-driven mutation entry extensions.

* **Chores**
  * Registered the new plugin and added its package manifest to the playground.

### DIFF
--- a/src/mutation-store.ts
+++ b/src/mutation-store.ts
@@ -174,25 +174,27 @@ export const useMutationCache = /* @__PURE__ */ defineStore(MUTATION_STORE_ID, (
       options: UseMutationOptionsWithDefaults<TData, TVars, TError, TContext>,
       key?: EntryKey | undefined,
       vars?: TVars,
-    ): UseMutationEntry<TData, TVars, TError, TContext> => extend(scope.run(() =>
-        markRaw<UseMutationEntry<TData, TVars, TError, TContext>>({
-          // only ids > 0 are real ids
-          id: 0,
-          state: shallowRef<DataState<TData, TError>>({
-            status: 'pending',
-            data: undefined,
-            error: null,
-          }),
-          gcTimeout: undefined,
-          asyncStatus: shallowRef<AsyncStatus>('idle'),
-          when: 0,
-          vars,
-          key,
-          options,
-          ext: {},
-        } satisfies UseMutationEntry<TData, TVars, TError, TContext>),
-      )!
-    ),
+    ): UseMutationEntry<TData, TVars, TError, TContext> =>
+      extend(
+        scope.run(() =>
+          markRaw<UseMutationEntry<TData, TVars, TError, TContext>>({
+            // only ids > 0 are real ids
+            id: 0,
+            state: shallowRef<DataState<TData, TError>>({
+              status: 'pending',
+              data: undefined,
+              error: null,
+            }),
+            gcTimeout: undefined,
+            asyncStatus: shallowRef<AsyncStatus>('idle'),
+            when: 0,
+            vars,
+            key,
+            options,
+            ext: {},
+          } satisfies UseMutationEntry<TData, TVars, TError, TContext>),
+        )!,
+      ),
   )
 
   /**
@@ -355,9 +357,7 @@ export const useMutationCache = /* @__PURE__ */ defineStore(MUTATION_STORE_ID, (
     TVars = unknown,
     TError = unknown,
     TContext extends Record<any, any> = _EmptyObject,
-  >(
-    entry: UseMutationEntry<TData, TVars, TError, TContext>,
-  ): Promise<TData> {
+  >(entry: UseMutationEntry<TData, TVars, TError, TContext>): Promise<TData> {
     // the vars is set when the entry is ensured, we warn against it below
     const { vars, options } = entry as typeof entry & { vars: TVars }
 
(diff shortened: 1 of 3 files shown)


src/pinia-colada.ts
```diff
@@ -69,10 +70,12 @@ export const PiniaColada: Plugin<[options?: PiniaColadaOptions]> = (
 
   // install plugins
   const queryCache = useQueryCache(pinia)
+  const mutationCache = useMutationCache(pinia)
   plugins?.forEach((plugin) =>
     plugin({
       scope: queryCache._s,
       queryCache,
+      mutationCache,
```
posva: We can’t have this because it removes the ability of tree shaking the mutations. The mustmanually call useMutationCache(pinia)

Barbapapazes: Oh whoo, wasn't thinking about it. I've changed it.

src/mutation-store.ts
```diff
@@ -353,12 +362,14 @@ export const useMutationCache = /* @__PURE__ */ defineStore(MUTATION_STORE_ID, (
    *
    * @param entry - the entry to mutate
    */
-  async function mutate<
+  const mutate = action(async <
```
posva: We don't need to wrap the action because it's not called within the mutation store

src/pinia-colada.spec.ts
```diff
@@ -47,6 +48,7 @@ describe('PiniaColada', () => {
     expect(plugin).toHaveBeenCalledTimes(1)
     expect(plugin).toHaveBeenCalledWith({
       queryCache: useQueryCache(pinia),
+      mutationCache: useMutationCache(pinia),
```
posva: this isn't true anymore, right?

src/use-mutation.spec.ts
```diff
@@ -121,6 +122,89 @@ describe('useMutation', () => {
...
+            [PiniaColada, { plugins: [PiniaColadaMutationMetricsTestPlugin()] }],
+          ],
+        },
+      )
+
+      // extensions should be available immediately (before first mutate)
+      expect(wrapper.vm.mutatedAt).toBe(0)
+      expect(wrapper.vm.errorCount).toBe(0)
+
+      vi.setSystemTime(1234)
```
posva: what is this meant to do? Is it `async vi.advanceTimersAsync(...)` that you wanted to use? What are you trying to trigger

posva (changes requested):
Nice! let's remove the new plugins/mutation-metricts. I understand it's nice to have an example but I think I prefer the docs only to not having to maintain. I rather point to one of your plugins as an example too

src/pinia-colada.spec.ts
Barbapapazes: yes, that's an artifact from the first version

docs/advanced/plugins.md
```diff
@@ -137,3 +170,43 @@ declare module '@pinia/colada' {
...
+import type { ShallowRef } from 'vue'
+
+export function PiniaColadaMutatedAtPlugin(): PiniaColadaPlugin {
+  return ({ pinia, scope }) => {
+    const mutationCache = useMutationCache(pinia)
+    mutationCache.$onAction(({ name, args, after }) => {
+      if (name === 'extend') {
+        const [entry] = args
+        scope.run(() => {
+          entry.ext.mutatedAt = shallowRef(0)
```
posva: I think you can reuse the plugins.md from the main branch (unless you added something new here)

Barbapapazes: yes, I removed it completely from this PR.

src/mutation-store.ts
```diff
@@ -158,28 +174,37 @@ export const useMutationCache = /* @__PURE__ */ defineStore(MUTATION_STORE_ID, (
...
-            key,
-            options,
-            // eslint-disable-next-line ts/ban-ts-comment
-            // @ts-ignore: some plugins are adding properties to the entry type
-            ext: START_EXT,
-          }) satisfies UseMutationEntry<TData, TVars, TError, TContext>,
-      )!,
+    ): UseMutationEntry<TData, TVars, TError, TContext> => {
+      const entry = scope.run(() =>
+        markRaw<UseMutationEntry<TData, TVars, TError, TContext>>({
```
posva: why markRaw?

src/mutation-store.ts
```diff
@@ -158,28 +174,37 @@ export const useMutationCache = /* @__PURE__ */ defineStore(MUTATION_STORE_ID, (
...
+          }),
+          gcTimeout: undefined,
+          asyncStatus: shallowRef<AsyncStatus>('idle'),
+          when: 0,
+          vars,
+          key,
+          options,
+          // eslint-disable-next-line ts/ban-ts-comment
+          // @ts-ignore: some plugins are adding properties to the entry type
+          ext: START_EXT,
```
posva: In this case, since we extend directly after, we can set to `{}` and call extend. Also let's just make `extend()` return the entry so we can keep the one liner and reduce size

src/mutation-store.ts
Barbapapazes: because this is what you've done in query-store.ts

posva: Ah haha, sorry about that

PR #495, opened 2026-02-09
docs: plugins

Barbapapazes:
## Summary by CodeRabbit

* **Documentation**
  * Rewrote plugin system docs: formal plugin structure, plugin context, action subscription patterns, and lifecycle guidance for queries and mutations.
  * Added practical plugin patterns, examples, and TypeScript augmentation guidance for extending query/mutation return values and global defaults.
  * Linked plugin guidance from query hooks, error handling, and mutations docs for better discoverability.

### DIFF
--- a/docs/advanced/plugins.md
+++ b/docs/advanced/plugins.md
@@ -17,9 +17,9 @@ Most plugin work falls into one of these categories:
 - observe outcomes centrally (logging, analytics, notifications)
 - coordinate timers/retries/refetch policies
 
-::: important
+::: tip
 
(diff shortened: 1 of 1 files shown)


Barbapapazes:
> Nice, this is a much needed addition Thanks for your tokens! 😁

The first commit has been written by AI but then I reorganize it to make sure it is complete, accessible and usable for both beginners and advanced users.

I wasn't sure it the first output would be usable and honestly, that was pretty good!

posva:
It's totally fine! I noticed becuase there were a few `::: info` that do not work in vitepress

PR #502, opened 2026-02-10
docs: improve plugins section in installation guide

Barbapapazes:


docs/guide/installation.md
```diff
@@ -99,4 +99,17 @@ app.use(PiniaColada, {
 })
 ```
 
-Other plugins must be installed separately (e.g. [Retries](https://github.com/posva/pinia-colada/tree/main/plugins/retry), [Loading delay](https://github.com/posva/pinia-colada/tree/main/plugins/delay), [Auto refetch](https://github.com/posva/pinia-colada/tree/main/plugins/auto-refetch)) and can also be created directly in your project for any custom behavior like offline, cache normalization, etc.
+Other plugins must be installed separately and registered the same way.
```
posva, line 102: Let's keep it as a one liner with a link to the official and community plugins

Issue #513, opened 2026-02-21
high gcTime leads to large memory heap size in SSR environments


dsfcoll:
I faced that data fetched via @pinia/colada is not being garbage collected after the request ends. This leads to a gradual increase in memory usage as more unique requests are made

Setting gcTime: 0 to solve the memory leak, but causes hydration mismatches

I expected gcTime to be ignored during SSR, or for the cache to be automatically scoped to the request lifecycle so that all cached data is garbage collected once the request is ended

After deploying a fix that set gcTime to 0 for SSR, the heap size droped.
<img width="1083" height="769" alt="Image" src="https://github.com/user-attachments/assets/1f187c70-9c7b-4147-8613-a6bda77efe4f" />

I also took heapsnapshot
<img width="1073" height="910" alt="Image" src="https://github.com/user-attachments/assets/96b68bd0-254a-4b4d-9166-2ab3192b9a35" />

As a workaround I left
`gcTime: import.meta.server ? false : staleTime`. It have less memory usage and there is no hydration mismatch errors

PR #514, opened 2026-02-21
fix(ssr): prevent memory leak

dsfcoll:
fixes https://github.com/posva/pinia-colada/issues/513

## Summary by CodeRabbit

* **New Features**
  * Added a configurable queryOptions setting that lets the plugin disable scheduled garbage collection during server-side rendering.

* **Bug Fixes**
  * Cancel pending garbage-collection timers when removing cached query entries to prevent lingering timeouts.
  * Ensure scheduled cache garbage collection is disabled on the server to avoid unnecessary cleanup during SSR.

### DIFF
--- a/nuxt/src/runtime/plugin.ts
+++ b/nuxt/src/runtime/plugin.ts
@@ -20,7 +20,7 @@ export default defineNuxtPlugin({
         ...coladaOptions.queryOptions,
         // Disable GC during SSR to prevent setTimeout closures from retaining
         // entry objects in memory across requests
-        ...(import.meta.server && { gcTime: false }),
+        gcTime: import.meta.server ? false : coladaOptions.queryOptions.gcTime,
       },
     } satisfies PiniaColadaOptions)
 


src/query-store.ts
```diff
@@ -895,11 +900,27 @@ export const useQueryCache = /* @__PURE__ */ defineStore(QUERY_STORE_ID, ({ acti
    * @param entry - the entry of the query to remove
    */
   const remove = action((entry: UseQueryEntry) => {
-    // setting without a value is like setting it to undefined
+    clearTimeout(entry.gcTimeout)
+    entry.gcTimeout = undefined
```
posva: we don't need to unset it since it should get gced anyway

src/query-store.ts
```diff
@@ -424,6 +424,11 @@ export const useQueryCache = /* @__PURE__ */ defineStore(QUERY_STORE_ID, ({ acti
     if (entry.deps.size > 0 || !entry.options) return
     clearTimeout(entry.gcTimeout)
     entry.pending?.abortController.abort()
+    // On the server, skip scheduling GC timeouts. The setTimeout closures
+    // would retain entry objects (and the entire Pinia store / Vue app via
+    // the closure chain) in memory until the timeout fires, causing a
+    // memory leak across SSR requests
+    if (!IS_CLIENT) return
```
posva: we should be able to safely remove this

nuxt/src/runtime/plugin.ts
```diff
@@ -27,7 +27,7 @@ export default defineNuxtPlugin({
           // it seems both mjs and cjs are included when doing a generate. This doesn't happen
           // in SSR. Clearing the cache after serialization avoids it being serialized twice, which
           // fails because query options are not serializable
-          queryCache.caches.clear()
+          queryCache.clear()
```
posva: instead of a custom action, we could just loop through all entries with something like `queryCache.getEntries().forEach(entry => queryCache.remove(entry))`. This will avoid adding a new action too

src/query-store.ts
dsfcoll: But the main idea is not to use setTimeout during SSR, because it retaining data even after the request has ended

posva: you can do that with a global `gcTime: isSSR ? false : ...`. In any case, it shouldn't be needed if `remove` clears the timeout

dsfcoll: I didn't know that `gcTime` could be set to false. We can just override `gcTime` to be false by default for SSR

`remove` is called only inside setTimeout, which doesn't actually solve the problem

To clarify the current situation:
For example we have 100 RPS and `gcTime` is set to 5 minutes, with each request fetching 50KB of data. After 5 minutes, the heap will contain approximately 100 * 60 * 5 * 50 KB = 1.5 GB of orphaned data and 30,000 active timeouts waiting to clear it

Overriding `gcTime` to false for SSR seems like a more elegant solution. If you agree, I will push the changes

posva: Sure, let's just set the gcTime to false in the nuxt module then

posva: although, clearing the timeout in remove is still a nice thing I think

dsfcoll: Pushed the changes

src/query-store.ts
```diff
@@ -896,7 +895,8 @@ export const useQueryCache = /* @__PURE__ */ defineStore(QUERY_STORE_ID, ({ acti
    * @param entry - the entry of the query to remove
    */
   const remove = action((entry: UseQueryEntry) => {
-    // setting without a value is like setting it to undefined
+    clearTimeout(entry.gcTimeout)
+    entry.gcTimeout = undefined
```
posva:
```suggestion
```

nuxt/src/runtime/plugin.ts
```diff
@@ -16,6 +16,12 @@ export default defineNuxtPlugin({
   setup(nuxtApp) {
     nuxtApp.vueApp.use(PiniaColada, {
       ...coladaOptions,
+      queryOptions: {
+        ...coladaOptions.queryOptions,
+        // Disable GC during SSR to prevent setTimeout closures from retaining
+        // entry objects in memory across requests
+        ...(import.meta.server && { gcTime: false }),
```
posva:
```suggestion
        gcTime: import.meta.server ? false : coladaOptions.queryOptions.gcTime,
```

Issue #536, opened 2026-03-24
queryCache.invalidateQueries does not work if enabled is initially false


J-Michalek:
# Description
invalidateQueries never refetches a query whose enabled option changed from false to true after the cache entry was first created. This is because `ensure()` doesn't update `entry.options` when the key hash is unchanged, so `invalidateQueries` reads a stale `enabled: false` from the cache entry.

## Reproduction
```ts
const open = ref(false)

const { state } = useQuery(() => ({
  key: ['my-query', id],
  query: () => fetchData(id),
  enabled: open.value,
}))

// Later: open the UI, query fires correctly via the watch(enabled) path
open.value = true

// Even later: try to invalidate while open is still true
const queryCache = useQueryCache()
await queryCache.invalidateQueries({ key: ['my-query', id] })
// ❌ No refetch happens
```
In `ensure()`, when the key hash matches the previous entry, it returns early without updating entry.options:

```ts
// query-store.ts — ensure()
if (previousEntry && keyHash === previousEntry.keyHash) return previousEntry;
//                                                      ^^^ early return

// entry.options = options  ← never reached
```
So if the entry was first created when enabled was false, `entry.options.enabled` stays false forever (as long as the key doesn't change).

## Workaround
Use the refresh returned by useQuery instead of `queryCache.invalidateQueries`:

```ts
const { refetch } = useQuery(() => ({ /* ... */ }))

// Instead of queryCache.invalidateQueries(...)
await refetch()
```

Version
@pinia/colada 1.0.0 (also present in 0.21.7)

posva:
Are you doing enabled: option or enabled: option.value ? Because the latter won’t work anyway

J-Michalek:
I am doing the getter function way.
```
useQuery(() -> ({ enabled: open.value }))
```

posva:
Ah yeah! Missed that! All good then

PR #537, opened 2026-03-25
fix: update entry options on ensure early return

babu-ch:
fixes https://github.com/posva/pinia-colada/issues/536

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where query options were not properly refreshed when reusing previously cached entries with matching keys, ensuring configurations remain consistent during query invalidation and updates.


Cat1007:
I want to know when this fix will be released. After updating to 1.0.0, a lot of logic that relies on invalidateQueries is invalid.

posva:
Done ✅

Issue #525, opened 2026-03-02
useInfiniteQuery throws "Cannot set properties of undefined" if the component of its first invocation is unmounted


vladyslav-mikhieiev:
#517  There is a bug in @pinia/colada introduced in version 0.21.6 (specifically with the PiniaColadaInfiniteQueryPlugin). 

If the very first invocation of useInfiniteQuery in the whole application happens inside a component's <script setup> or setup(), the plugin attaches its $onAction subscription to the queryCache store. Because detached: true is not passed to $onAction inside the plugin, this setup mistakenly binds the listener to the active component's EffectScope.

When the user navigates away and this component gets unmounted, Pinia cleans up and destroys the $onAction listener.

However, installedMap.set(queryCache, true) was already called when the plugin ran for the first time. If the user navigates to a new page (or navigates back), the next useInfiniteQuery call will see that installedMap.has(queryCache) is true, so it will not re-attach the $onAction listener.

Since the listener is dead, it never creates nextPageParam or previousPageParam inside entry.ext. When computePageParams subsequently runs, it attempts to assign values to exts.nextPageParam.value, resulting in the following Unhandled Error:

```js
TypeError: Cannot set properties of undefined (setting 'value')
    at computePageParams (chunk-XXXX.js:xxx)
    at useInfiniteQuery (chunk-XXXX.js:xxx)
```

<img width="1253" height="130" alt="Image" src="https://github.com/user-attachments/assets/7746d685-fcec-429e-94fe-93d7e4b75e31" />

**Steps to Reproduce**:

1. Do not use useInfiniteQuery globally or at the App root level.
2. Navigate to Page A where useInfiniteQuery is used. The component successfully fetches data.
3. Navigate to Page B where useInfiniteQuery is NOT used. Page A is unmounted, and the $onAction listener is destroyed.
4. Navigate back to Page A (or to any other page that uses useInfiniteQuery).
5. The App throws a TypeError inside computePageParams because entry.ext.nextPageParam is undefined.

**Expected behavior**: The PiniaColadaInfiniteQueryPlugin should attach the $onAction listener with detached: true to prevent it from being bound to any specific component's lifecycle:

```js
queryCache.$onAction(({ name, args }) => {
  // ...
}, true /* detached */)
```

**Temporary Workaround**: As a temporary fix, developers can initialize a dummy useInfiniteQuery call in the root 

App.vue
 script setup or in a global boot file where the EffectScope will not be unmounted during the user session:

```js
[...]

PR #549, opened 2026-04-10
fix(infinite-query): guard computePageParams against missing ext refs

tzzo:
We're hitting `TypeError: Cannot set properties of undefined (setting 'value')` in production (Nuxt SSR app, ~55 events across 35 users on v1.0.0/v1.1.0). The crash is in `computePageParams` when `entry.ext.nextPageParam` is undefined.

#525 fixed the `$onAction` detachment, but if `scope.run()` doesn't execute for any reason (e.g. stopped EffectScope), the ext refs are never created and `computePageParams` blows up.

This adds a guard that lazily initializes the extensions using the existing `createInfiniteQueryEntryExtensions` helper when they're missing. Includes a test that reproduces the broken state and verifies recovery.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented crashes when restoring paginated cached data missing pagination metadata, improving stability during hydration of saved data.

* **Tests**
  * Added coverage to verify preseeded paginated cache is recognized on mount and pagination state (pages/page params and next-page availability) is preserved.

### DIFF
--- a/src/infinite-query.spec.ts
+++ b/src/infinite-query.spec.ts
@@ -1582,4 +1582,31 @@ describe('useInfiniteQuery', () => {
     expect(w2.vm.data).toEqual({ pages: [[1, 2, 3]], pageParams: [0] })
     expect(w2.vm.hasNextPage).toBe(true)
   })
+
+  it('does not crash when entry ext is missing nextPageParam', async () => {
+    const pinia = createPinia()
+    const app = createApp({})
+    app.use(pinia)
+    app.use(PiniaColada)
+    const queryCache = useQueryCache(pinia)
+
+    // simulate a cache entry whose ext refs were never initialized
+    // (e.g. the plugin's scope.run() didn't execute)
+    const entry = queryCache.ensure({
+      key: ['key'],
+      query: async () => ({ pages: [[1, 2, 3]], pageParams: [0] }),
+      meta: { __i: true },
+    })
+    entry.ext = {} as any
+
+    // mounting should not throw
+    const { wrapper } = mountSimple({ staleTime: 1000 }, { plugins: [pinia] })
+
+    await flushPromises()
+
+    expect(wrapper.vm.data).toEqual({
+      pages: [[1, 2, 3]],
+      pageParams: [0],
+    })
+  })
 })
--- a/src/infinite-query.spec.ts
+++ b/src/infinite-query.spec.ts
@@ -1474,6 +1474,30 @@ describe('useInfiniteQuery', () => {
       // After everything settles, hasNextPage should remain true
       expect(wrapper.vm.hasNextPage).toBe(true)
     })
+
+    it('can set the data before using infinite queries', async () => {
+      const pinia = createPiniawithHydratedCache({
+        // '["key"]': [{ pages: [[1, 2, 3]], pageParams: [0] }, null, 0, { __i: true }],
+      })
+      const queryCache = useQueryCache(pinia)
+      queryCache.setQueryData(['key'], { pages: [[1, 2, 3]], pageParams: [0] })
+
+      const { wrapper } = mountSimple(
+        {
+          staleTime: 1000,
+          getNextPageParam: (_lastPage, _allPages, lastPageParam) =>
+            lastPageParam >= 3 ? null : lastPageParam + 1,
+        },
+        { plugins: [pinia] },
+      )
+      await flushPromises()
+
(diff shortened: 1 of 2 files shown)


src/infinite-query.ts
```diff
@@ -418,6 +418,10 @@ export function useInfiniteQuery<
     if (!entry) return
     const lastPageParam = data?.pageParams.at(-1)
     const exts = entry.ext as unknown as UseInfiniteQueryExtensions<TPageParam>
+    // extensions might not be initialized if the plugin's scope was stopped
+    if (!exts.nextPageParam) {
+      createInfiniteQueryEntryExtensions(exts as unknown as UseInfiniteQueryExtensions<unknown>)
+    }
```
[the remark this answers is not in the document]
tzoulas: Fixed

posva (changes requested):
The test passes without the fix. Can you investigate further and ensure the test fails without the fix first to validate it as a simplified version of the reproduction?

posva:
Sorry, left the review on the wrong tab!

src/infinite-query.spec.ts
```diff
@@ -1582,4 +1582,31 @@ describe('useInfiniteQuery', () => {
...
+
+  it('does not crash when entry ext is missing nextPageParam', async () => {
+    const pinia = createPinia()
+    const app = createApp({})
+    app.use(pinia)
+    app.use(PiniaColada)
+    const queryCache = useQueryCache(pinia)
+
+    // simulate a cache entry whose ext refs were never initialized
+    // (e.g. the plugin's scope.run() didn't execute)
```
posva: I need to understand why this would happen. There might be a hidden bug. I tried to reproduce by hydrating the store but that doesn't create the bug. We need a realistic test scenario (or a reliable reproduction) to move forward

posva (changes requested):

src/infinite-query.spec.ts
posva: oh actually, nevermind, I ran some tests again and I think I found the bug

posva: Okay, I dug deeper and I confirm my suspicious. I need an actual reproduction for this. I tried reproducing it with

```ts
    it('does not crash when ext refs were not initialized (stopped scope)', async () => {
      const pinia = createPiniawithHydratedCache({
        '["key"]': [{ pages: [[1, 2, 3]], pageParams: [0] }, null, 0, { __i: true }],
      })
      const queryCache = useQueryCache(pinia)

      const { wrapper } = mountSimple(
        {
          staleTime: 1000,
          getNextPageParam: (_lastPage, _allPages, lastPageParam) =>
            lastPageParam >= 3 ? null : lastPageParam + 1,
        },
        { plugins: [pinia] },
      )
      await flushPromises()

      expect(wrapper.vm.data).toEqual({
        pages: [[1, 2, 3]],
        pageParams: [0],
      })
      expect(wrapper.vm.hasNextPage).toBe(true)
    })
  })
```

but that just works because `useInfiniteQuery()` adds the `PiniaColadaInfiniteQueryPlugin` and ensures the entry, which calls `extend()`.

It's as if in your codebase the infinite query was being _ensured_ (and therefore extended) before `useInfiniteQuery()` is called, and that's the source of the bug

tzzo: Could you please try this: https://github.com/tzzo/pinia-colada-repro-549? I struggle to get this working with codesandbox or StackBlitz :/. 

Steps:
  1. pnpm install && pnpm dev
  2. Click "Pre-populate cache" (calls queryCache.setQueryData(key, { pages, pageParams }))
  3. Navigate to /list (mounts useInfiniteQuery for that key) → crash

This happens when optimistic mutations (adding an item to a list for example) call setQueryData() with InfiniteData-shaped values on pages where the corresponding useInfiniteQuery() hasn't mounted yet.

posva: I see it know. I think we need special methods to treat infinite queries, as separate exports to not add to the bundle, and a warning in the console to point at the mistake if made.

You will have to write something like:


```ts
import { setInfiniteQueryData } from '@pinia/colada'

setInfiniteQueryData(queryCache, key, data)
```

As a workaround, maybe this works:

- Call `useInfiniteQuery()` (even with fake data) to initialize the Infinite Queries plugin (not exposed)
- Before calling `setQueryData()`, create and ensure the entry manually with `{meta: {__i: 1}}`(not 100% about the feasibility here)

PR #562, opened 2026-04-20
fix(cache-persister): await captured promise on async restore

Danny-Devs:
The async restore path at `plugins/cache-persister/src/cache-persister.ts:181` awaits a fresh `storage.getItem(key)` call instead of the already-captured `storedPromise`, causing two reads on every app boot when storage is async (IndexedDB, AsyncStorage, remote KV, etc.). Sync storage (localStorage) is unaffected.

Introduced in 0f536f5 ("fix: avoid extra tick") — the variable name `storedPromise` and the in-code comment both show the original intent was to reuse the captured promise; the async branch just missed it.

### Change

```diff
- const stored = storedPromise instanceof Promise ? await storage.getItem(key) : storedPromise
+ const stored = storedPromise instanceof Promise ? await storedPromise : storedPromise
```

### Test

Adds a call-count assertion to the existing async-storage test block. Without the fix, `getItem` is called twice on restore; with the fix, once.

---

## Summary by CodeRabbit

* **Tests**
  * Added new test case for async-storage cache restoration to validate that cached data is properly retrieved from persistent storage and restored correctly on initialization

* **Refactor**
  * Improved cache restoration efficiency by streamlining the underlying storage read logic to prevent redundant operations and unnecessary storage requests during cache rehydration processes

### DIFF
--- a/plugins/cache-persister/src/cache-persister.spec.ts
+++ b/plugins/cache-persister/src/cache-persister.spec.ts
@@ -430,6 +430,19 @@ describe('PiniaColadaCachePersister', () => {
       expect(queryCache.getQueryData(['async-restored'])).toBe('async-cached')
     })
 
+    it('calls getItem once when restoring from async storage', async () => {
+      const storage = createAsyncStorage()
+      storage.data['pinia-colada-cache'] = JSON.stringify({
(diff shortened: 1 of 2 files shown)


posva (approved):
Thanks!

PR #563, opened 2026-04-20
fix(retry): prevent retryMap key collisions for structurally distinct keys

Danny-Devs:
## Summary

`plugins/retry` uses `entry.key.join('/')` as its `retryMap` key, which collides for structurally distinct keys — silently cancelling sibling retries when colliding queries fail around the same time.

## The bug

`Array.prototype.join('/')` is lossy:
- `['a/b'].join('/')` === `['a', 'b'].join('/')` === `"a/b"`
- `['users', {id: 1}]` and `['users', {id: 2}]` both coerce to `"users/[object Object]"`

The cache disambiguates these via `keyHash`; the retry plugin does not. When a colliding query's retry fires, the internal `queryCache.fetch()` re-enters the `'fetch'` handler and runs `clearTimeout` on the shared `retryMap` entry — cancelling the sibling's scheduled retry.

## Fix

Swap `entry.key.join('/')` → `entry.keyHash` at the `'remove'` and `'fetch'` handlers. `keyHash` is already public on `UseQueryEntry` and uses the same canonical serialization the cache keys entries with.

## Regression tests

Two tests under `retryMap key uniqueness`, each mounting colliding-key pairs and asserting independent retry cycles:

- `['a/b']` vs `['a', 'b']`
- `['users', {id: 1}]` vs `['users', {id: 2}]`

Both fail on the buggy code and pass on the fix.

## Test plan

- [x] `pnpm -F @pinia/colada-plugin-retry exec vitest run` — 23/23 pass
- [x] Full library suite — 498 pass, 11 todo, 0 fail
- [x] `oxlint` clean
- [x] Two-sided verification (revert → 2 tests fail, restore → all pass)

---

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive test coverage for independent retry scheduling across multiple simultaneous queries, verifying that different query keys correctly maintain separate retry scheduling state.

* **Refactor**
  * Updated retry bookkeeping mechanism to use consistent key identification across all retry operations, ensuring reliable handling of retry scheduling and cleanup.

### DIFF
--- a/plugins/retry/src/retry.spec.ts
+++ b/plugins/retry/src/retry.spec.ts
@@ -530,6 +530,89 @@ describe('Pinia Colada Retry Plugin', () => {
     })
   })
 
+  describe('retryMap key uniqueness', () => {
+    function twoQueryFactory(a: UseQueryOptions, b: UseQueryOptions) {
+      const wrapper = mount(
+        defineComponent({
+          template: '<div></div>',
+          setup() {
+            const qa = useQuery(a)
+            const qb = useQuery(b)
+            return { qa, qb }
+          },
+        }),
+        {
+          global: {
+            plugins: [
+              createPinia(),
+              [PiniaColada, { plugins: [PiniaColadaRetry(RETRY_OPTIONS_DEFAULTS)] }],
+            ],
+          },
+        },
+      )
+      return { wrapper }
+    }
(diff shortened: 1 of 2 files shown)


posva (approved):
Thanks!

PR #568, opened 2026-04-23
fix(retry): clear scheduled retry on cancel

Danny-Devs:
## Summary

`plugins/retry` doesn't respond to the `cancel` action. After a failed query, the plugin schedules a `setTimeout` for the next attempt. If the user calls `queryCache.cancel(entry)` between retries, the timer keeps ticking — when it fires, it dispatches a new fetch against the user's explicit cancel.

## The bug

The `$onAction` subscriber handles `extend`, `remove`, and `fetch`. `cancel` is missing.

1. Fetch fails → `onError(retryFetch)` schedules `setTimeout(..., delay)`, stores `timeoutId` in `retryMap`, sets `isRetrying=true` / `retryCount=N` / `retryError=e` on `entry.ext`.
2. No in-flight request between retries — just the timer.
3. User calls `queryCache.cancel(entry)`. `cancel` in `src/query-store.ts:754` aborts `entry.pending` (nothing pending) and nulls it. It does not touch `retryMap` or the plugin's `ext` refs.
4. Timer fires. Guard at `retry.ts:164` checks `entry.active` and `options.enabled` — both still truthy, `cancel` doesn't deactivate. Proceeds to `queryCache.fetch(queryEntry)` — the zombie request.

Symptom: the UI shows the cancel as "taken," then a mystery network call fires `delay` ms later and may even update the query's data.

## Fix

Add a `cancel` handler that mirrors `remove`'s `retryMap` cleanup, plus resets the reactive UI refs. Unlike `remove`, the cache entry survives `cancel`, so the `isRetrying` / `retryCount` / `retryError` refs — still observed by any mounted component — need to be reset, otherwise the UI continues to show "retrying..." indefinitely.

## Regression tests

Two tests under `cancel behavior`:

- `does not fire the scheduled retry after queryCache.cancel()` — asserts the query fn is called exactly once after cancel, even after advancing timers past the retry window.
- `resets retry UI flags on cancel` — asserts `isRetrying=false`, `retryCount=0`, `retryError=null` after cancel.

Both fail on the pre-fix code and pass with the fix.

The `factory` helper now also returns `pinia` so tests can resolve `useQueryCache(pinia)` and call `cancel` on the entry. Existing tests using `const { wrapper } = factory(...)` are unaffected.

## Test plan

- [x] `pnpm -F @pinia/colada-plugin-retry exec vitest run` — 25/25 pass
- [x] Full library suite — 501 pass, 11 todo, 0 fail
- [x] `oxlint` clean (no new warnings)
- [x] `oxfmt --check` clean
- [x] Typecheck clean
- [x] Two-sided verification: revert fix → both new tests fail; restore → all pass

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Cancelling or removing queries now clears scheduled retries and immediately resets all retry-related state (retry count, retry status, and error), preventing pending retries from running.

* **Tests**
  * Added tests that verify cancellation by key and by entry prevents scheduled retries and immediately resets retry-related UI/state flags.

### DIFF
--- a/plugins/retry/src/retry.ts
+++ b/plugins/retry/src/retry.ts
@@ -82,13 +82,15 @@ export function PiniaColadaRetry(
           entry.ext.retryError = shallowRef(null)
         })
         if (process.env.NODE_ENV === 'development') {
-          entry.ext.retry = { isRetrying: false, retryCount: 0, retryError: null as unknown }
+          entry.ext.retry = {
+            isRetrying: false,
+            retryCount: 0,
+            retryError: null,
+          }
         }
         return
-      }
-
-      // cleanup all pending retries when data is deleted (means the data is not needed anymore)
-      if (name === 'remove') {
+      } else if (name === 'remove' || name === 'cancel') {
+        // cleanup all pending retries
         const [cacheEntry] = args
         const key = cacheEntry.keyHash
         const entry = retryMap.get(key)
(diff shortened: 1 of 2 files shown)


plugins/retry/src/retry.ts
```diff
@@ -98,6 +98,25 @@ export function PiniaColadaRetry(
...
+        const key = cacheEntry.keyHash
+        const entry = retryMap.get(key)
+        if (entry) {
+          clearTimeout(entry.timeoutId)
+          retryMap.delete(key)
+        }
+        cacheEntry.ext.isRetrying.value = false
+        cacheEntry.ext.retryCount.value = 0
+        cacheEntry.ext.retryError.value = null
+        if (process.env.NODE_ENV === 'development' && cacheEntry.ext.retry) {
```
posva, line 106: Why this? It seems out of scope

Danny-Devs:
Great question. The reasoning behind it, briefly: I'd convinced myself `isRetrying` could end up out of sync with reality in several places across the retry lifecycle, and I tacked the reset block onto the cancel handler to patch them all. Re-reading retry.ts, your plugin already handles those cases correctly (L217-219 on success, L207-208 on max-retries, L184-192 on inactive-while-timer-pending, L142-151 on manual refetch).

The only place `isRetrying` genuinely drifts — where `<Spinner v-if="isRetrying">` would keep showing after the retry cycle has ended — is the cancel path:

- `isRetrying` is set `true` at L171 together with scheduling the timer at L182.
- `queryCache.cancel()` in query-store.ts only calls `abort()`, sets `asyncStatus = 'idle'`, and nulls `entry.pending` — it doesn't touch retry refs.
- None of the existing reset sites are reachable from cancel: L142-151 is gated on `name === 'fetch'`, L184-192 only fires if the timer isn't cleared, L207 and L217 only fire when retryFetch runs on a new error or success — cancel produces neither.

So `isRetrying` stays `true` after cancel until the user triggers another fetch. I was trying to fix that together with adding the cancel handler itself, and I picked the wrong pattern — zeroing `retryCount`, which contradicts your L207 treatment that deliberately keeps it as a record. Apologies for mixing the two concerns into one block.

Four possible directions:

**1. Idiomatic** — mirror your L207 pattern: reset `isRetrying` + `retryError`, keep `retryCount`. Worth flagging honestly — this is still a variant of the block you just questioned. It's only the right answer if your "why this?" was about the pattern I chose (too aggressive), not about ref-touching in cancel generally. If the latter, this option doesn't fit.

**2. Structural (computed)** — make `isRetrying` a computed over `retryMap`. Since `retryMap.has(key)` is already the single source of truth for "is a retry scheduled," deriving `isRetrying` from it eliminates the whole drift class. Every existing `retryMap.delete(key)` site (remove, manual refetch, timer-inactive, max-retries, success, cancel) automatically flips `isRetrying` false — no imperative resets anywhere. Net change is subtractive — the existing imperative `isRetrying.value = …` writes at L144, L171, L185, L207, L217 become dead code (roughly 5-8 lines removable). Requires making retryMap `shallowReactive` so the computed tracks it — standard Vue pattern.

**3. API-level** — drop `isRetrying` entirely and let consumers compose (TanStack's approach: `isFetching && failureCount > 0`). Probably not a fit here though: TanStack keeps `fetchStatus: 'fetching'` throughout a retry cycle including the delay windows between attempts, so that composition captures the wait window. Pinia Colada's `asyncStatus` returns to `'idle'` during the delay between retries, so an analogous composition wouldn't cover that window — dropping `isRetrying` would lose information nothing else carries. Mentioning for completeness.

**4. Just revert, skip the drift discussion** — keep this PR strictly to timer + map cleanup mirroring `remove`, and leave the drift question for a separate issue (or not at all if you don't think it's worth addressing).

My read: Option 2 is the cleanest — it treats the imperative ref-management as the actual issue and replaces it with derivation over your existing source of truth. Option 1 is smaller but doesn't change the underlying pattern.

Let me know which direction you prefer. If Option 1, I can reshape the block in this PR. If Option 2 or 3, I can revert this PR and open a follow-up. If you'd rather I bundle Option 2 or 3 into this PR, just let me know.

posva:
I dont mis if you use AI to drive thé PR but please dont just submit what it says

Danny-Devs:
Sorry--I know that was a really long comment. However I vetted/validated the whole thing and edited it multiple times to make sure it is correct/accurate, especially making sure the logic is sound.

If any of it is not on target or doesn't make sense, please let me know.

I will try to keep things more concise next time.

posva:
thanks. Yes, I would appreciate a more concise wording

Danny-Devs:
More concise version:
Your question about the block was good.

I was originally trying to patch `isRetrying` drift on the `cancel` path, but the drift only occurs in one path (I thought it occurred on more paths). The path that it occurs in: `isRetrying` stays true after cancel, because none of the reset code is reachable from there.

Two solutions I thought were good:
1) The `remove` path already has reset code that the `cancel` path can copy.

2) Use `computed` to solve the `isRetrying` problem. `retryMap.has(key)` is the single source of truth for "is a retry scheduled," so deriving `isRetrying` from it eliminates the whole drift class. However, this would involve making `retryMap` a `shallowReactive`, which might not be what you want.

I can remove the block, or I can do 1) or 2) if you want this PR to address the drift issue as well.

posva (approved):
Thanks!

PR #570, opened 2026-04-23
fix(query-store): avoid false HMR invalidation when the same component mounts multiple times

### LINKED ISSUE #569: HMR id tracking causes redundant refetches when the same component mounts across multiple ticks (dev-only)
## Summary

In development mode, `useQuery` triggers a fresh fetch for every new instance of the **same component** when those instances mount in separate ticks, even though they share an identical query key. In production this does not happen and the query is deduplicated as expected.

## Environment

- `@pinia/colada`: 1.2.0
- `pinia`: latest
- `vue`: 3.x
- Vite dev server (development build)

## Reproduction

Minimal repro — no third-party components, just Vue + pinia + pinia-colada.

### `Child.vue`

```vue
<script lang="ts" setup>
import { useQuery } from '@pinia/colada'

let fetchCount = 0

const { data, isLoading } = useQuery({
  key: () => ['shared-items'],
  query: async () => {
    fetchCount++
    console.log('[useQuery] fetch called, total fetches =', fetchCount)
    await new Promise(r => setTimeout(r, 1000))
    return ['a', 'b', 'c']
  },
})
</script>

<template>
  <div>query: {{ isLoading ? 'loading...' : data?.join(',') }}</div>
</template>
```

### `Parent.vue`

```vue
<script lang="ts" setup>
import { onMounted, ref } from 'vue'
import Child from './Child.vue'

const items = ref<number[]>([])

onMounted(async () => {
  for (let i = 0; i < 7; i++) {
    await new Promise(r => setTimeout(r, 10))
    items.value.push(i)
  }
})
</script>

<template>
  <div v-for="i in items" :key="i">
    <Child />
  </div>
</template>
```

### main.ts
```ts
import { createApp } from 'vue'
import Parent from './Parent.vue'
import { createPinia } from 'pinia'
import { PiniaColada } from '@pinia/colada'

createApp(App)
  .use(createPinia())
  .use(PiniaColada, {
    queryOptions: {
      refetchOnMount: false,
      refetchOnWindowFocus: false,
    }
  })
  .mount('#app')

```
## Expected behavior

All 7 `<Child />` instances share the same query key `['shared-items']`, therefore only **one** `fetch` should be called and every instance should subscribe to the same pending promise.

## Actual behavior

In **development**: 7 fetches are triggered — one per mounted instance. Each new mount cancels the previous one.

```
[useQuery] fetch called, total fetches = 1
[useQuery] fetch called, total fetches = 2
...
[useQuery] fetch called, total fetches = 7
```

In **production** (`vite build && vite preview`): only 1 fetch, as expected.

## Root cause

In `src/query-store.ts` `ensure()` contains a dev-only block that treats the appearance of the same component `__hmrId` on a cache entry as a signal that the component's code was hot-updated, and calls `invalidate(entry)`:

```js
if (process.env.NODE_ENV !== "production") {
  const currentInstance = getCurrentInstance();
  if (currentInstance) {
    entry.__hmr ??= { ids: new Map() };
    const id = currentInstance.type?.__hmrId;
    if (id) {
      if (entry.__hmr.ids.has(id)) invalidate(entry); // <-- fires on a legitimate 2nd instance
      const count = (entry.__hmr.ids.get(id) ?? 0) + 1;
      entry.__hmr.ids.set(id, count);
    }
  }
}
```

This heuristic does not distinguish between:

1. A hot-module replacement that legitimately produced a new `setupFn` for the same `__hmrId` (the intended case).
2. Multiple instances of the same component being mounted (the common case: lists, virtualized tables, tabs, accordions, infinite scroll, etc.).

Whenever case (2) happens across ticks, each new mount invalidates the shared cache entry and aborts the in-flight request, forcing a refetch.

When all instances mount in the **same tick** the bug is hidden: `ensure()` runs N times synchronously with no pending request yet, and subsequent `refresh()` calls in `onMounted` find the first fetch already pending and reuse its promise. As soon as mounts happen across ticks (which is normal in lists with async data, virtualized grids, transitions, etc.), every new instance cancels the previous request.

## Impact

- Any UI that renders the same component multiple times with staggered mounting (virtualized tables like ag-grid, infinite scroll, staged transitions, async lists) multiplies network requests in development.
- Masks real dedup behavior during development, making it hard to notice when the real request pattern is wrong in production.
- Behavior differs between dev and prod, which is itself surprising.

## Suggested fix

The HMR heuristic needs to detect an actual module update rather than a new instance. Options:

- Track identity of the resolved `setupFn` / component options (reference equality) rather than `__hmrId`. A real HMR swap produces a new object; a new instance reuses the same one.
- Hook into Vite's `import.meta.hot` directly for the owning module and invalidate on the `accept`/`update` event, not on mount.
- At minimum, guard the invalidation behind a check that the count is `> 1` only after an HMR event has been observed for that id.

vad1ym:
Closes #569.

## Summary

In dev, `queryCache.ensure` invalidated a cache entry whenever the same `__hmrId` registered against it — which also fires when the same component is legitimately mounted multiple times (lists, virtualized grids, staged mounts). The existing `FIXME` / `TODO` comments in `src/query-store.ts` already described this exact problem and hinted at the fix.

Distinguish a real HMR update from a new instance by tracking `type.render`: HMR swaps the reference, a new instance keeps it.

## Changes

- `src/query-store.ts` — `__hmr.ids` stores `{ count, render }`; `invalidate(entry)` fires only when `render` reference changes.
- `src/use-query.spec.ts`:
  - Replaced `it('always refetches on hmr', ...)` — it asserted the old buggy behavior. Now simulates a real code swap (new `render`, same `__hmrId`) and still asserts 2 fetches.
  - Added a test for the repro from #569: 7 staggered mounts → 1 fetch.

## Verification

- New test fails on `main` (7 calls), passes with the fix (1 call).
- `use-query` + `query-store` + `define-query`: 163 passed, no regressions.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Hot Module Replacement detection to properly distinguish between actual code updates and new component instances, ensuring queries are only refetched when necessary.
  * Enhanced query request deduplication for multiple simultaneous component mounts of the same component, reducing redundant fetches and improving performance in list/grid scenarios.

### DIFF
--- a/src/query-store.ts
+++ b/src/query-store.ts
@@ -143,9 +143,13 @@ export interface UseQueryEntry<
    */
   __hmr?: {
     /**
-     * Reference count of the components using this query.
+     * Per-hmrId tracking: reference count of mounted components and the last seen
+     * `render` function. A real HMR update replaces the component's `render`
+     * reference, so a changed reference signals an actual code update (and
+     * warrants invalidation), while an unchanged reference means a legitimate
+     * new instance of the same component (no invalidation needed).
      */
-    ids: Map<string, number>
+    ids: Map<string, { count: number, render: unknown }>
   }
 }
 
@@ -403,11 +407,15 @@ export const useQueryCache = /* @__PURE__ */ defineStore(QUERY_STORE_ID, ({ acti
     // clear any HMR to avoid letting the set grow
     if (process.env.NODE_ENV !== 'production') {
       if ('type' in effect && '__hmrId' in effect.type && entry.__hmr) {
-        const count = (entry.__hmr.ids.get(effect.type.__hmrId) ?? 1) - 1
-        if (count > 0) {
-          entry.__hmr.ids.set(effect.type.__hmrId, count)
+        const existing = entry.__hmr.ids.get(effect.type.__hmrId as string)
+        const count = (existing?.count ?? 1) - 1
+        if (count > 0 && existing) {
+          entry.__hmr.ids.set(effect.type.__hmrId as string, {
+            count,
+            render: existing.render,
+          })
         } else {
-          entry.__hmr.ids.delete(effect.type.__hmrId)
+          entry.__hmr.ids.delete(effect.type.__hmrId as string)
         }
       }
     }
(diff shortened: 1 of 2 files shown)


posva (commented):
Thanks for the PR! After taking a look, I realized it needed more changes and new tests so I opened #571

PR #578, opened 2026-05-02
refactor!: use unjs/c12 as config loader

kilisei:
This PR moves the old configuration loader to 'unjs/c12' and introduces support for the '.config/' directory, as well as a more uniform configuration file name and structure.

There is a breaking change in this PR because we no longer use the '.options' file instead, we use the '.config' file. 

This PR adds one dependency: c12 v4-beta, which is ~400 KB big.

## Summary by CodeRabbit

* **New Features**
  * Optional watch mode for Colada config in development with live reload and change logging
  * Runtime-loaded, strongly typed Colada config support and updated playground demonstrating plugin hook registration

* **Chores**
  * Added a new runtime dependency
  * Restructured config loading and updated module exports/types; runtime now resolves config at load time

* **Documentation**
  * Nuxt docs updated to show the new colada block, watch option, and config examples


posva:
Thanks but I rather not add an extra dependency for something that already works. Adding support for a `.config/` folder is not something I'm interested in. That plus the breaking change, plus the size increase, I really don't think it's worth it...

Why did you want to add it?

kilisei:
> Thanks but I rather not add an extra dependency for something that already works. Adding support for a `.config/` folder is not something I'm interested in. That plus the breaking change, plus the size increase, I really don't think it's worth it...
> 
> Why did you want to add it?

I mostly wanted to add it for the watch option and more uniform config names (Nuxt supports .config and nuxt.config). I have all my configs in the .config directory. I find it cleaner and less messy than having ten config files in the root directory, but I understand why you won't merge it. 🙂

I could make it backwards compatible if you would be interested to that.

Bundle size increase is 17.7% + 0.83kB

posva:
The name being .options is, IMO, important here. It like the router.options.ts file
So let’s drop this ideas, at least for the moment. Thanks for the PR though!

PR #590, opened 2026-05-17
fix(infinite-query): store nextPageIndicator on entry.ext instead of closure

jasonlewis:
As I brought up in #589, I believe this could be a bug. I took a swing at it with a new test where I was able to reproduce the issue.

Essentially, the local next page indicator causes issues when the same query is mounted more than once. I've instead moved that indicator onto the cache entry so it can be shared between queries with the same key.

Cheers.

## Release Notes

* **Tests**
  * Added test coverage to verify pagination continues correctly when multiple components share the same query cache.

* **Refactor**
  * Improved internal state tracking for multi-page query loading to ensure consistent behavior across shared component instances.

[Review Change Stack](https://app.coderabbit.ai/change-stack/posva/pinia-colada/pull/590?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/infinite-query.spec.ts
+++ b/src/infinite-query.spec.ts
@@ -658,6 +658,106 @@ describe('useInfiniteQuery', () => {
     })
   })
 
+  it('loadNextPage continues from the last page across multiple components sharing the same key', async () => {
+    const pinia = createPinia()
+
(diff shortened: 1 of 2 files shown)


posva (approved):
Thanks, nice catch! Any state that is attached to the query should indeed be in the entry itself to be reused properly.

PR #615, opened 2026-06-30
fix(devtools): normalize component resolver path on Windows

thyngster:
The unplugin-vue-components resolver returned a raw resolve() path, which on Windows produces backslashes (I:\pinia-colada\...). Vite's import-analysis cannot resolve those as import specifiers, causing "Failed to resolve import" for the U* custom-element components. Wrap the path in normalizePath() to emit forward slashes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved component loading in the developer tools on Windows by handling file paths more consistently.
  * Reduced import resolution issues that could prevent `.ce.vue` components from being processed correctly.

### DIFF
--- a/devtools/vite.config.ts
+++ b/devtools/vite.config.ts
@@ -1,6 +1,6 @@
 import { dirname, resolve } from 'node:path'
 import { fileURLToPath } from 'node:url'
-import { defineConfig } from 'vite'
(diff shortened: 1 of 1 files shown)


posva (approved):
Thanks!

PR #629, opened 2026-08-08
fix(persister): account for time spent in storage on restore

### LINKED ISSUE #627: Cache-persister: restored snapshots are treated as fresh regardless of how long they sat in storage
**Plugin version:** `@pinia/colada-plugin-cache-persister@1.1.0`
**Pinia Colada version:** `@pinia/colada@1.4.2`

## Problem

The persister serializes entries via `_queryEntry_toJSON`, which stores each entry's age **relative to serialization time** (`Date.now() - when`). That format is designed for SSR, where serialize and hydrate happen within the same request. For long-lived persistence it discards the wall-clock time the snapshot spent in storage: on `hydrateQueryCache`, `when` is reconstructed as `Date.now() - age`, so a snapshot persisted seconds after a fetch is always restored as "seconds old" — no matter whether it was written a minute or a month ago.

## Impact

A snapshot persisted on Friday and restored on Monday is served as fresh for the full `staleTime` — no refetch on mount. Users see days-old data on app start. We hit this in production with `staleTime: 10min` + IndexedDB storage: customers opening the app after a weekend saw Friday's data for up to 10 minutes, with no refetch being triggered.

## Reproduction

1. Set up `useQuery` with `staleTime: 600_000` and the cache persister with any storage.
2. Let the query fetch and the persister write its snapshot.
3. Close the app; reopen it days later. (In a test: use fake timers and advance the clock past `staleTime` **between** persist and restore — advance the system time only, without firing pending timers.)
4. On mount the query does **not** refetch; `entry.stale` is `false` for the next ~10 minutes even though the data is days old.

```ts
// condensed test (vitest, fake timers)
const storage = createMemoryStorage();

// session 1: fetch + persist
const app1 = mountQueryApp(storage, query1); // useQuery({ key, query, staleTime: 600_000 })
await flushPromises();
vi.advanceTimersByTime(200); // fire persister debounce
await flushPromises();
app1.unmount();

// a weekend passes
vi.setSystemTime(Date.now() + 3 * 24 * 60 * 60 * 1000);

// session 2: restore
resetCacheReady();
const app2 = mountQueryApp(storage, query2);
await isCacheReady();
await flushPromises();

expect(query2).toHaveBeenCalledTimes(1); // ❌ fails: never called, snapshot served as fresh
```

## Suggested fix

Wrap the persisted payload with an absolute timestamp and add the elapsed time back to each entry age on restore:

```ts
// stringify
JSON.stringify({ persistedAt: Date.now(), entries })

// parse
const elapsed = Math.max(0, Date.now() - persistedAt);
// for each entry [data, error, age, meta]: age >= 0 ? age + elapsed : age
```

This keeps the instant first paint (data is still hydrated) while making restored entries correctly stale, so they refetch on first use.

We currently do exactly this in our app via the `stringify`/`parse` options as a workaround.

fudianchn:
## Problem

`@pinia/colada-plugin-cache-persister` serializes each entry's age **relative to serialize time** (`Date.now() - when`, via `_queryEntry_toJSON`). That format is designed for same-tick SSR serialize/hydrate, but for **long-lived storage** (e.g. IndexedDB) the wall-clock time the snapshot spends in storage is lost: `hydrateQueryCache` reconstructs `when = Date.now() - age`, so a snapshot persisted seconds after a fetch is restored as fresh no matter whether it was written a minute or a month ago.

**Impact:** with `staleTime` + persistent storage, a snapshot persisted on Friday and restored on Monday is served as fresh for the full `staleTime` — users see days-old data on app start with no refetch. (#627 reports this from production.)

## Fix

Persist an **absolute** fetch timestamp (`entry.when`) instead of the relative age, and on restore recompute the relative age `hydrateQueryCache` expects from the time elapsed since the snapshot was written (`Date.now() - when`). The relative encoding in core (`_queryEntry_toJSON` / `hydrateQueryCache`) is untouched — the plugin converts to/from absolute at its own boundary.

### Backward compatibility

Snapshots persisted by older versions stored a small relative age. Absolute epoch-ms timestamps are `>= 1e12` (post-2001), so anything smaller is recognized as a legacy relative age and hydrated unchanged — **no behavior change for existing persisted data**. The `-1` "no timestamp" sentinel is also left untouched.

## Testing

Added a regression test that reproduces the issue end to end:
- session 1: fetch + persist at `T0` (fake clock), then unmount;
- advance the system clock well past `staleTime` (simulating time spent in storage);
- session 2: restore from the same storage and assert the entry is now stale, so the query refetches on mount.

Verified **fail-before / pass-after** the change, and the full plugin suite is green (32/32), including the existing `Date`-round-trip-with-devalue and pre-populated-restore tests, which confirm backward compatibility is preserved.

```
 ✓ src/cache-persister.spec.ts (32 tests) 
 Test Files  1 passed (1)
      Tests  32 passed (32)
```

Formatted with `oxfmt`; linted with `oxlint`; `tsgo` type-check clean.

## Alternatives considered

Per the discussion in #627, an alternative is to wrap the persisted payload as `{ persistedAt, entries }` and add the elapsed time on restore. I went with storing the absolute timestamp in the existing entry slot because it needs **no change to the persisted shape or to the `stringify`/`parse` contract** (custom codecs such as `devalue` are unaffected) and no extra storage entry, while still being fully backward compatible. Happy to switch to the wrapped-payload approach if that's preferred.

Closes #627

## Summary by CodeRabbit

* **Bug Fixes**
  * Restored persisted cache entries now correctly account for the time elapsed since they were fetched.
  * Stale data is refetched after restoration instead of being treated as fresh.
  * Existing persisted cache snapshots remain compatible with the updated restoration behavior.

* **Tests**
  * Added coverage to verify cache expiration across extended periods between sessions.

### DIFF
--- a/plugins/cache-persister/src/cache-persister.spec.ts
+++ b/plugins/cache-persister/src/cache-persister.spec.ts
@@ -251,6 +251,33 @@ describe('PiniaColadaCachePersister', () => {
       expect(useQueryCache().getQueryData(['restored'])).toBe('restored-data')
     })
 
+    it('treats restored entries as stale by the time they spent in storage', async () => {
+      const storage = createMockStorage()
+      const staleTime = 60_000
+      const query = vi.fn(async () => 'cached')
+
+      // session 1: fetch and persist at T0
+      vi.setSystemTime(new Date(2026, 0, 1, 0, 0, 0))
+      const { wrapper: app1 } = factory(
+        { key: ['weekend'], query, staleTime },
+        { storage, debounce: 0 },
+      )
+      await runPersist()
+      expect(query).toHaveBeenCalledTimes(1)
+      app1.unmount()
+
+      // the snapshot sits in storage well past staleTime; only the clock moves
+      vi.setSystemTime(new Date(2026, 0, 4, 0, 0, 0))
+
+      // session 2: restore — the entry is days old, so it must refetch on mount
+      resetCacheReady()
+      factory({ key: ['weekend'], query, staleTime }, { storage })
+      await flushPromises()
+      await flushPromises()
+
+      expect(query).toHaveBeenCalledTimes(2)
+    })
+
     it('isCacheReady resolves after restore', async () => {
       const storage = createMockStorage()
       storage.data['pinia-colada-cache'] = JSON.stringify({
--- a/plugins/cache-persister/src/cache-persister.ts
+++ b/plugins/cache-persister/src/cache-persister.ts
@@ -182,7 +182,17 @@ export function PiniaColadaCachePersister(
               queryCache
                 // we only care about entries with data
                 .getEntries({ ...filter, status: 'success' })
-                .map((entry) => [entry.keyHash, _queryEntry_toJSON(entry)]),
+                .map((entry) => {
+                  const serialized = _queryEntry_toJSON(entry)
+                  // Persist an absolute fetch timestamp. _queryEntry_toJSON stores
+                  // the entry age relative to serialize time, which only holds
+                  // while serialize and hydrate run in the same tick (SSR). For
+                  // long-lived storage the wall-clock time the snapshot spends in
+                  // storage would otherwise be lost and restored entries would be
+                  // treated as fresh no matter how old they are (see restore()).
+                  if (entry.when) serialized[2] = entry.when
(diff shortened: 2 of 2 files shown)


posva:
Thanks for the PR, I went a different way #630

